PR #12, opened 2025-09-10
Burnwire

ineskhou:
#### Description

This adds Burnwire component into the topology

###  How has this been tested?

- [ ] Unit tests
- [X] Integration tests
- [ ] Z Tests
- [X] Manual testing: Ran the burnwire, felt it get hot and tested different safety timers 

## Checklist

- [X] Written detailed sdd with requirements, channels, ports, commands, telemetry defined and correctly formatted and spelled
- [X] Have written relevant integration tests and have documented them in the sdd
- [X] Have done a code review with

### DIFF
--- a/FprimeZephyrReference/Components/Burnwire/Burnwire.cpp
+++ b/FprimeZephyrReference/Components/Burnwire/Burnwire.cpp
@@ -29,10 +29,6 @@ void Burnwire ::burnStop_handler(FwIndexType portNum) {
     this->stopBurn();
 }
 
-// void Burnwire ::stop_handler(FwIndexType portNum) {
-//     //TODO
-// }
-
 void Burnwire::startBurn() {
     this->log_ACTIVITY_HI_SetBurnwireState(Fw::On::ON);
     this->m_safetyCounter = 0;
--- a/FprimeZephyrReference/Components/Burnwire/docs/sdd.md
+++ b/FprimeZephyrReference/Components/Burnwire/docs/sdd.md
@@ -59,4 +59,4 @@ Add unit test descriptions in the chart below
 ## Parameter
 | Name | Description |
 | -----|-------------|
-|   SAFETY_TIMER   | By Default set in fpp (currently 3) is the max time the burnwire should ever run|
+|   SAFETY_TIMER   | By Default set in fpp (currently 10) is the max time the burnwire should ever run|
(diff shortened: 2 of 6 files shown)


FprimeZephyrReference/test/int/burnwire_test.py
```diff
@@ -0,0 +1,83 @@
...
+def reset_burnwire(fprime_test_api: IntegrationTestAPI):
+    """Fixture to stop burnwire and clear histories before/after each test"""
+    # Stop burnwire and clear before test
+    fprime_test_api.send_and_assert_command(
+        f"{COMPONENT}.STOP_BURNWIRE", [], max_delay=2
+    )
+    fprime_test_api.clear_histories()
+    yield
+    # Clear again after test to prevent residue
+    fprime_test_api.clear_histories()
```
nateinaction, line 28: Do you also want to stop the burnwire here?

nateinaction (commented):
When the burn starts, it might be nice to add the amount of time the burn will last for to the event. On burn end it might be nice to add an arg to that event too showing how long the burn was. I think adding these args to the events could help ground ops and integration tests.

Is there a way for you to try and set the param in the integration test? If you set the timeout param to a really long amount, this could help you show that the second manual case is actually stopping early rather than timing out. This could also give you a third usecase to see if changing the param actually impacts the timeout time.

FprimeZephyrReference/test/int/burnwire_test.py
ineskhou: Good idea

FprimeZephyrReference/Components/Burnwire/Burnwire.fpp
```diff
@@ -0,0 +1,88 @@
...
+
+        sync command STOP_BURNWIRE(
+        )
+
+
+        # @ Example telemetry counter
+        # telemetry ExampleCounter: U64
+
+        # @ Example event
+        # event ExampleStateEvent(example_state: Fw.On) severity activity high id 0 format "State set to {}"
```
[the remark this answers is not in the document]
ineskhou: The ReferenceDeployment.burnwire.SafetyTimerState does emit to say how long the burn will be at the begining, but I will add an event that says how long the burnwire actually ended up burning for! More data is better

As for setting parameters in the tests, the reason I didn't do this is because parameters are stored in the filesystem and my understanding is that we don't have ours setup yet. I was getting errrors changing the paramater

PR #30, opened 2025-10-01
Prune zephyr manifest

Mikefly123:
# Improving CI Build Times

## Description
This pull request adds a custom `west.yml` Zephyr manifest and also refactors the `build` CI to significantly reduce the GitHub Actions Runtime! 

By my best estimate we are going to put that build time in about half! 
Screenshot 2025-10-03 at 1 49 16 PM

## How Has This Been Tested?

- [X] Ran on GitHub Actions

## Checklist


nateinaction (approved):
I _think_ the zephyr module downloads are not being cached anymore because, at least locally, are being downloaded to a parent directory of the repo. They are indeed faster but and the builds are functional so I think this is an OK step forward but we should pend more time here to understand how to store the modules in the repo instead.

Issue #73, opened 2025-11-01
[FEAT] Distance Sensor


yudataguy:
The distance sensor is [VL6180](https://www.st.com/en/imaging-and-photonics-solutions/vl6180.html#overview) Time-of-Flight (ToF) proximity sensor. 

Based on @Mikefly123 suggestion:
1. Use Zephyr TCA to create a general reader to read the sensor data
2. Pass the data to Fprime
3. Write a Fprime component to do something with the data

Reference:

[Use devices' drivers behind I2C mux](https://github.com/zephyrproject-rtos/zephyr/discussions/57027)

Please correct me.

PR #77, opened 2025-11-04
Distance Sensor

yudataguy:
#73 Adding distance sensor, due to hardware issue, this is only a draft for now, until investigation is done

### DIFF
--- a/FprimeZephyrReference/project/config/CMakeLists.txt
+++ b/FprimeZephyrReference/project/config/CMakeLists.txt
@@ -5,6 +5,7 @@
 ####
 register_fprime_config(
     CONFIGURATION_OVERRIDES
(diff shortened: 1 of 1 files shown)


Mikefly123:
I'm going to close this for now as descoped and not important to work on. 

One day we might return to trying to use the distance sensor or we might just remove it from the kit for simplicity. The original purpose of the distance sensor was to provide closed loop control for the relatively unreliable burn wire circuit in the V1 and V2 PROVES Kit. With the new V5 Flight Controller board in the V3 PROVES Kit the burnwire circuit is now significantly safer and more reliable than previous iterations, negating the need for the closed loop control. 

During the final checkouts of Alcyone, Atlas, and Electra we found that the antennas would consistently deploy using a 10s burn timer (usually deploying in around 6-8s in air and within 2-3s in low vacuum.

PR #79, opened 2025-11-04
Introduce Detumble Component

hrfarmer:
# Detumble Component

## Description

This PR introduces the Detumble component for the core reference.

- 💤 Detumble is `DISABLED` by default and should be enabled by the startup sequence.
- ✏️ All math formulas are recorded in comments in hpp files for future contributors and are unit tested.
- 👨‍🔬 Tested using multimeter, ocilliscope, internal power consumption metrics, and a magnetometer.
- 🦺 Safe mode automatically disables detumble
- 🧪 Added support for debugger in VS Code

Once `detumbleManager.OPERATING_MODE_PRM_SET` is set to `AUTO` the `detumbleManager.State` will cycle between `COOLDOWN` -> `SENSING` -> `TORQUING` -> `COOLDOWN` whenever the satellite is tumbling. After the satellite has stabilized, the state will remain in `SENSING`. The detumble system telemetry can be viewed by setting the telemetry level to `6`. 

### Readings with magnetometer
[Data sheet](https://docs.google.com/spreadsheets/d/1a0DMlIpKav0Lw4UmIbdproPtMVpxzjvouQ32-nclAiI/edit?gid=547742505#gid=547742505)

Setup with magnetometer ~1inch from center of XY face.
IMG_7194

Readings in microtesla after setting `drv2605Face0Manager.START` to `127`, `-127, and `0`.
Captura de pantalla 2025-12-23 a la(s) 19 18 52

### Readings with osciliscope
Readings after setting `drv2605Face0Manager.START` to `127`.
IMG_7197

### Functional demonstration
Showing telemetry from GDS after turning on all faces, setting `detumbleManager.OPERATING_MODE_PRM_SET` to `AUTO`, setting `bDotDetumble.GAIN_PRM_SET` to `5`, and manually bringing CalPoly 2U above rotational threshold.

[Captura de pantalla 2025-12-23 a la(s) 20 09 50](https://youtu.be/mbYkjTt2x7I)

### Automated DRV2605 test
DRV2605 integration test now works since power draw has been increased
Captura de pantalla 2025-12-23 a la(s) 20 49 33

### Automated unit tests
`BDot`, `Magnetorquer`, and `StrategySelector` helper classes all have automated tests
Captura de pantalla 2025-12-28 a la(s) 00 10 03

### DIFF
--- a/FprimeZephyrReference/Components/ImuManager/ImuManager.cpp
+++ b/FprimeZephyrReference/Components/ImuManager/ImuManager.cpp
@@ -137,7 +137,7 @@ Drv::MagneticField ImuManager ::magneticFieldGet_handler(FwIndexType portNum, Fw
 
     if (!device_is_ready(this->m_lis2mdl)) {
         this->log_WARNING_HI_Lis2mdlDeviceNotReady();
-        return Drv::MagneticField(0.0, 0.0, 0.0, Fw::ZERO_TIME);
+        return Drv::MagneticField(0.0, 0.0, 0.0, Fw::TimeValue());
     }
     this->log_WARNING_HI_Lis2mdlDeviceNotReady_ThrottleClear();
 


hrfarmer:
This should be the majority of what's needed for detumble, but won't work until the magnetorquer driver is done.

FprimeZephyrReference/test/int/camera_handler_test.py
```diff
@@ -9,7 +9,7 @@
 
 import pytest
 from common import proves_send_and_assert_command
-from fprime.common.models.serialize.time_type import TimeType
+from fprime_gds.common.models.serialize.time_type import TimeType
```
nateinaction: Fix for deprecation notice
```sh
FprimeZephyrReference/test/int/drv2605_test.py:12
  /Users/nate/code/github.com/open-source-space-foundation/proves-core-reference/FprimeZephyrReference/test/int/drv2605_test.py:12: DeprecationWarning: Ground type system has been migrated to the fprime-gds package, in fprime_gds.common.models.serialize. Change your imports accordingly.
    from fprime.common.models.serialize.time_type import TimeType
```

FprimeZephyrReference/test/int/drv2605_test.py
```diff
@@ -9,7 +9,7 @@
 
 import pytest
 from common import proves_send_and_assert_command
-from fprime.common.models.serialize.time_type import TimeType
+from fprime_gds.common.models.serialize.time_type import TimeType
```
nateinaction: Fix for deprecation notice

FprimeZephyrReference/test/int/power_monitor_test.py
```diff
@@ -8,8 +8,8 @@
 
 import pytest
 from common import proves_send_and_assert_command
-from fprime.common.models.serialize.time_type import TimeType
 from fprime_gds.common.data_types.ch_data import ChData
+from fprime_gds.common.models.serialize.time_type import TimeType
```
nateinaction: Fix for deprecation notice

FprimeZephyrReference/Components/Drv/Drv2605Manager/Drv2605Manager.hpp
```diff
@@ -129,6 +126,8 @@ class Drv2605Manager final : public Drv2605ManagerComponentBase {
     //! Continuous mode
     //! If true, the magnetorquer will be triggered on every run port call
     bool m_continuous_mode = false;
+    bool m_has_initialized = false;
+    int m_count = 0;
```
ineskhou: Are these being used or is this from when we had contious mode? Just waneed to check when we check mags are initilized if so

FprimeZephyrReference/Components/DetumbleManager/DetumbleManager.cpp
```diff
@@ -0,0 +1,349 @@
...
+
+    // Clamp to max current
+    if (std::fabs(targetCurrent) > maxCurrent) {
+        clampedCurrent = (targetCurrent > 0) ? maxCurrent : -maxCurrent;
+    } else {
+        clampedCurrent = targetCurrent;
+    }
+
+    // Scale to int8_t range [-127, 127]
+    return static_cast<I8>(std::round((clampedCurrent / maxCurrent) * 127.0));
```
ineskhou: Super edge case that shouldnt happen unless coil resistance is 0 which shouldnt happen but could maybe have a /0 here so maybe want to check

ineskhou (approved):
WOW!!! This is so incredibly impressive, you guys really thought of a lot of edge cases and operator uses and a lot of the questions I had were answered in sdds or notes or just clear code. I added a few small nitpicks but no pressure to look at those and wait for Michaels reivwe b4 merging but yea!! THANK YOU SO MUCH!!!!

(also some of the integtation testsa are failing (most I think bc of the int test setup w/o sensors and stuff, but jsut check that theres nothing unexpected since I know Saidi fixed those up)

hrfarmer (commented):
I can't explicitly request changes since its technically my pr, but this should wait to go in until #237 & #239 are in as well.

(Haven't reviewed everything yet, will be able to get to it very soon)

FprimeZephyrReference/Components/DetumbleManager/DetumbleManager.cpp
```diff
@@ -0,0 +1,349 @@
...
+    if (condition != Fw::Success::SUCCESS) {
+        this->log_WARNING_LO_AngularVelocityRetrievalFailed();
+        return;
+    }
+    this->log_WARNING_LO_AngularVelocityRetrievalFailed_ThrottleClear();
+
+    F64 angular_velocity_magnitude = this->getAngularVelocityMagnitude(angular_velocity);
+
+    // If angular velocity is below threshold, remain in SENSING state
+    // TODO(nateinaction): Consider transitioning the mode to DISABLED if below threshold
```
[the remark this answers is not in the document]
nateinaction: Yeah, I think it's worth talking with the team about the behavior of the component as written and if we want the behavior to be different. After writing and using the component I think it is safe to leave in AUTO as long as you don't have any other magnetorquer control systems running. Any thoughts @Mikefly123?

FprimeZephyrReference/Components/Drv/Drv2605Manager/Drv2605Manager.hpp
nateinaction: Ah and I found a few more to clean up and some comments in this file that needed fixing. Thanks!

FprimeZephyrReference/Components/DetumbleManager/DetumbleManager.cpp
```diff
@@ -0,0 +1,349 @@
...
+    DetumbleMode mode = this->paramGet_OPERATING_MODE(isValid);
+
+    // Telemeter mode
+    this->tlmWrite_Mode(mode);
+
+    // Telemeter state
+    this->tlmWrite_State(this->m_detumbleState);
+
+    // If detumble is disabled, ensure magnetorquers are off and exit early
+    if (this->paramGet_OPERATING_MODE(isValid) == DetumbleMode::DISABLED) {
```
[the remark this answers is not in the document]
nateinaction: Good catch, the paramGet lookup contains a mutex lock/unlock so good to reduce.

FprimeZephyrReference/Components/Drv/Drv2605Manager/Drv2605Manager.cpp
```diff
@@ -12,6 +12,13 @@
 
 namespace Drv {
 
+static uint8_t input_arr[1];
+static struct drv2605_rtp_data rtp = {
+    .size = 1,
+    .rtp_hold_us = (uint32_t[]){1},
```
Mikefly123: I am a little confused at setting the rtp_hold_us to only 1us here. I may just be confused and this is overwritten somewhere else, but I couldn't figure it out in a few minutes of spelunking. Does this configuration make it so the rtp runs constantly until it is told to stop?

Mikefly123 (approved):
Hey @hrfarmer and @nateinaction great work on this PR! Really awesome to see that we're able to stick the landing on this one right before handoff. Great job documenting the in lab test cases in the summary too! 

Overall I think things in here look good and I would be onboard with merging this in with two main caveats that I think we should try and address with follow up PR's or discussion and deferment before handoff:

1. Because the detumbling is a power negative action (potentially _really_ power negative as you have seen on the 2U during testing) we ideally would want to add a port that allows the `DISABLED` mode to be automagically called when we enter `SAFE_MODE` due to a low battery condition. This would potentially save our bacon in the unfortunate case that a satellite encounters both a high tip off rotation _and_ we are unable to command it from ground to stop burning power on detumbling.

2. Per one of my comments I think we should try and implement some controller dead bands both around the velocity threshold and in the case of extremely high angular velocity. I think this is especially important because we are not doing any filtering on our sensor data, so we might have a hard time settling to a steady state without a dead band.

Re: the thought about combining the detumble and bdot component, I actually think it is good to keep the math and the operation in separate components. At least when I was reading things it made sense to me and it doesn't feel like too much additional overhead.

FprimeZephyrReference/Components/Drv/Drv2605Manager/Drv2605Manager.cpp
nateinaction: Yes, that's right. This value causes the `haptics_start_output()` call to block for that amount of time but the RTP configuration keeps running until the input value is `rtp_input` value is set to 0. For our purposes, the lower the `rtp_hold_us` value the faster we can cycle this component. I will add comments to clarify this in the code.

tests/CMakeLists.txt
```diff
@@ -5,16 +5,55 @@ enable_testing()
...
+    get_filename_component(test_name ${test_src} NAME_WE)
+
+    add_executable(${test_name} ${test_src})
+    target_link_libraries(${test_name}
+        gtest_main
+        detumble_manager_magnetorquer
+        detumble_manager_strategy_selector
+        detumble_manager_bdot
+        rtc_manager_rtc_helper
+    )
```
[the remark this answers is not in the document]
nateinaction: This was an intentional change to make it less cumbersome to maintain these tests... It's easy to split back out if we think this suggestion is worth implementing.

FprimeZephyrReference/Components/DetumbleManager/StrategySelector.cpp
```diff
@@ -0,0 +1,82 @@
...
+}
+
+void StrategySelector ::configure(double bdot_max_threshold,
+                                  double deadband_upper_threshold,
+                                  double deadband_lower_threshold) {
+    this->m_bdot_max_threshold = bdot_max_threshold;
+    this->m_deadband_upper_threshold = deadband_upper_threshold;
+    this->m_deadband_lower_threshold = deadband_lower_threshold;
+
+    if (this->m_rotation_target != deadband_upper_threshold || this->m_rotation_target != deadband_lower_threshold) {
```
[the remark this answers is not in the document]
nateinaction: The intention is to prevent the target from not being one of either the `deadband_upper_threshold` or `deadband_lower_threshold`. Will add a comment.

FprimeZephyrReference/Components/Drv/Drv2605Manager/Drv2605Manager.hpp
```diff
@@ -121,14 +118,6 @@ class Drv2605Manager final : public Drv2605ManagerComponentBase {
...
     Fw::Time m_load_switch_on_timeout;
-
-    //! Load switch check
-    //! Available to disable if the component is not powered by a load switch
-    bool m_load_switch_check = true;
-
-    //! Continuous mode
-    //! If true, the magnetorquer will be triggered on every run port call
-    bool m_continuous_mode = false;
 };
```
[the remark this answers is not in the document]
nateinaction: This does not appear to be accurate.

FprimeZephyrReference/Components/DetumbleManager/DetumbleManager.cpp
```diff
@@ -0,0 +1,400 @@
...
+    // Perform torqueing action
+    this->startMagnetorquers(this->m_x_plus_magnetorquer.dipoleMomentToCurrent(dipole_moment[0]),
+                             this->m_x_minus_magnetorquer.dipoleMomentToCurrent(dipole_moment[0]),
+                             this->m_y_plus_magnetorquer.dipoleMomentToCurrent(dipole_moment[1]),
+                             this->m_y_minus_magnetorquer.dipoleMomentToCurrent(dipole_moment[1]),
+                             this->m_z_minus_magnetorquer.dipoleMomentToCurrent(dipole_moment[2]));
+}
+
+void DetumbleManager ::hysteresisTorqueAction() {
+    // Perform torqueing action
```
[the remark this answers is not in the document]
nateinaction: There is now a hysteresis axis selection parameter.

FprimeZephyrReference/Components/DetumbleManager/BDot.cpp
```diff
@@ -0,0 +1,127 @@
...
+    if (std::chrono::milliseconds(10) > dt_us) {
+        // Out of order readings or readings taken too quickly, cannot compute dipole moment
+        errno = EINVAL;
+        return false;
+    }
+
+    // TODO(nateinaction): Take this in as an argument
+    // TODO(nateinaction): The 500ms limitation needs to be recorded in documentation, this prevents detumble from
+    // working on slower rate groups Set previous magnetic field if last reading time was more than 500ms ago
+    if (dt_us > std::chrono::milliseconds(500)) {
```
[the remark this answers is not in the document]
nateinaction: I agree.

### ODR / Minimum Reading Delta
For taking in the ODR, this is a little tricky to do since the `BDot` class cannot reach out to the `Drv2605Manager` directly. We could:
1. Create an ODR enum in the `BDot` class that mirrors the ODR enum from the `Drv2605Manager` component. 
2. Use compile time validation in the `DetumbleManager` component constructor to ensure the enums are the same.
3. Create an input port in the `Drv2605Manager` component to get the current ODR from the parameter
4. Create an output port in the `DetumbleManager` component to get the ODR
5. Create a method in the `BDot` class to allow the ODR to be configured from `DetumbleManager`
6. Create a method in the `BDot` class to determine how fast to set the minimum reading delta from the ODR

This would cause the `BDot` class to have to know about the ODRs of the specific magnetometer we're using which is not great. Instead we could create an input port in the `Drv2605Manager` component to get the minimum time between readings. That feels cleaner.

### Rate Group / Maximum Reading Delta

FprimeZephyrReference/ReferenceDeployment/Top/ReferenceDeploymentPackets.fppi
```diff
@@ -168,6 +168,76 @@ telemetry packets ReferenceDeploymentPackets {
...
+  packet Detumble id 15 group 6 {
+    ReferenceDeployment.detumbleManager.Mode
+    ReferenceDeployment.detumbleManager.State
+    ReferenceDeployment.detumbleManager.DetumbleStrategy
+  }
+
+  packet DetumblePerformance id 16 group 6 {
+    ReferenceDeployment.detumbleManager.TorqueDuration
+    ReferenceDeployment.detumbleManager.TimeBetweenMagneticFieldReadings
+  }
```
Mikefly123: I think you can make the argument that these should be in `Group 3 - Satellite Meta Data`. Although I also see the counter argument that this telemetry either only needs to be requested once (`id15`) or is just for debug (`id16`). We could designate a new group that is just for debug telemetry.

FprimeZephyrReference/Components/DetumbleManager/DetumbleManager.cpp
```diff
@@ -0,0 +1,432 @@
...
+    // Transition to SENSING state
+    this->m_state = DetumbleState::SENSING;
+}
+
+void DetumbleManager ::stateSensingActions() {
+    // Get angular velocity thresholds from parameters
+    Fw::ParamValid isValid;
+    F64 bdot_max_threshold = this->paramGet_BDOT_MAX_THRESHOLD(isValid);
+    F64 deadband_upper_threshold = this->paramGet_DEADBAND_UPPER_THRESHOLD(isValid);
+    F64 deadband_lower_threshold = this->paramGet_DEADBAND_LOWER_THRESHOLD(isValid);
```
Mikefly123: Hey @nateinaction since I know you're very interested in writing code that would be functionally safe for human spaceflight here's a little something that isn't critical, but could be good to think about. 

Because the thresholds here are being set as three separate parameters than than a single parameter that gets unpacked into 3 values, there is a possibility that a ground operator error or a dropped packet could cause only one of the values to not update correctly when you intended to update all multiple values simultaneously. 

Your strategy selection and the bdot algorithm don't actually have many ways of going wrong, so it's low consequence for this component. The worst think I could think of is maybe we want to floor the `MaxThreshold` in flight to force hysteresis mode, but the strategy selector keeps choosing `IDLE` because the `lower_threshold` is higher than the new `MaxThreshold`. Not a bad situation to be in really, and it can be fixed easily on the next ground station pass.

I have heard of an example of a spacecraft that apparently experienced a partial loss of vehicle due to incomplete (partial) setting of software safety thresholds on a control system in flight though. That's a bit of an extreme case, but if you wanted to flex some functional safety thinking to introduce a software pattern, here would be a nice place to do it! 

One way to do it could be to force all parameters to update simultaneously to ensure the operator can't just update one and forget to do another. Another thing you may want to do is validate assumptions, like `LOWER_THRESHOLD` < `UPPER_THRESHOLD` < `MAX_THRESHOLD`.

Mikefly123 (dismissed):
Great work on updating this Nate! Love the `StrategySelector`. Once again, I'll put the LGTM on this and say it looks good to merge in when we are ready! Maybe we can review as a team on Tuesday?

My comments are 1 further thinking suggestion, 1 recommendation for operator understanding, and 1 nitpick.

FprimeZephyrReference/Components/DetumbleManager/docs/sdd.md
```diff
@@ -0,0 +1,320 @@
...
+    }
+    StrategySelector -- Strategy
+```
+
+#### BDot Maximum Threshold
+
+The B-Dot maximum threshold parameter defines the upper rotational rate (deg/s) above which the hysteresis strategy is used instead of B-Dot. This threshold is computed as:
+
+$$
+\omega_{\max} = \min\!\left(\frac{2\pi}{\Delta t}, \frac{\pi}{2\,\delta T}\right)
```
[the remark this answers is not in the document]
nateinaction: Review with Evan:
We should try to reduce the effect of noise in magnetometer readings by taking more than 2 measurements.

FprimeZephyrReference/Components/DetumbleManager/BDot.hpp
```diff
@@ -0,0 +1,87 @@
...
+    // ----------------------------------------------------------------------
+
+    // Compute the required dipole moment to detumble the satellite.
+    //
+    // m = -k * (dB/dt) / |B|
+    //
+    // m is the dipole moment in A⋅m²
+    // k is a gain constant
+    // dB/dt is the time derivative of the magnetic field reading in micro-Tesla per second (uT/s)
+    // |B| is the magnitude of the magnetic field vector in micro-Tesla (uT)
```
nateinaction: Review with Evan:
Double check units here. Magnetometer reading comments say that it's in gauss

FprimeZephyrReference/Components/DetumbleManager/DetumbleManager.cpp
```diff
@@ -0,0 +1,432 @@
...
+                return;
+        }
+    }
+    this->log_WARNING_LO_MagneticFieldTooSmallForDipoleMoment_ThrottleClear();
+    this->log_WARNING_LO_InvalidMagneticFieldReadingForDipoleMoment_ThrottleClear();
+    this->log_WARNING_LO_UnknownDipoleMomentComputationError_ThrottleClear();
+
+    // Perform torqueing action
+    this->startMagnetorquers(this->m_x_plus_magnetorquer.dipoleMomentToCurrent(dipole_moment[0]),
+                             this->m_x_minus_magnetorquer.dipoleMomentToCurrent(dipole_moment[0]),
```
nateinaction: Review with Evan:
Does powering both magnitorquers in axis overcompensate the requested current? YES! Oops.

FprimeZephyrReference/Components/DetumbleManager/DetumbleManager.fpp
```diff
@@ -0,0 +1,337 @@
...
+        param BDOT_MAX_THRESHOLD: F64 default 150.0 id 40
+
+        @ Parameter for storing the upper deadband rotational threshold in deg/s
+        param DEADBAND_UPPER_THRESHOLD: F64 default 8.0 id 41
+
+        @ Parameter for storing the lower deadband rotational threshold in deg/s, below which detumble is considered complete
+        param DEADBAND_LOWER_THRESHOLD: F64 default 5.0 id 1
+
+        @ Parameter for storing the cooldown duration
+        param COOLDOWN_DURATION: Fw.TimeIntervalValue default {seconds = 0, useconds = 20000} id 3
```
nateinaction: Document how cooldown duration was determined.

FprimeZephyrReference/Components/DetumbleManager/DetumbleManager.cpp
```diff
@@ -0,0 +1,432 @@
...
+            return;
+        case DetumbleState::TORQUING:
+            this->stateTorquingActions();
+            return;
+    }
+}
+
+void DetumbleManager ::setMode_handler(FwIndexType portNum, const Components::DetumbleMode& mode) {
+    this->m_mode = mode;
+}
```
nateinaction: Maybe don't allow detumble to turn on if in safe mode?

FprimeZephyrReference/Components/DetumbleManager/DetumbleManager.fpp
```diff
@@ -0,0 +1,337 @@
...
+        ### Parameters ###
+
+        @ Parameter for storing the upper rotational threshold in deg/s, above which bdot detumbling is replaced by hysteresis detumbling. Given by ω_max = min(2π/∆t, π/2δT) where ∆t is the duration of actuation and δT is the time elapsed between the measurement of Ḃ.
+        param BDOT_MAX_THRESHOLD: F64 default 150.0 id 40
+
+        @ Parameter for storing the upper deadband rotational threshold in deg/s
+        param DEADBAND_UPPER_THRESHOLD: F64 default 8.0 id 41
+
+        @ Parameter for storing the lower deadband rotational threshold in deg/s, below which detumble is considered complete
+        param DEADBAND_LOWER_THRESHOLD: F64 default 5.0 id 1
```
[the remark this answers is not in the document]
Mikefly123: I think a default of 1 degree per second is too slow to go to without also having an algorithm that will attempt to orient the satellite in a given direction (like MOVE-II did). The risk incurred is that the satellite may settle in a poor orientation (like antenna board directly pointed at the sun) or suffer thermal issues due to now barbecue rolling to balance out the heat taken on from the sun. 

Rather than adding a timer to the component I think you can just extend the startup sequence to say that after 3 hours the detumble mode is forced to `DISABLE`.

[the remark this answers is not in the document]
nateinaction: Thanks on both suggestions.

Note to myself.
Need to document decision to stay at $5\ \deg/s$

FprimeZephyrReference/Components/DetumbleManager/BDot.hpp
nateinaction: Completed this check with @evanjellison

FprimeZephyrReference/Components/DetumbleManager/BDot.cpp
```diff
@@ -0,0 +1,98 @@
...
+std::array<double, 3> BDot ::computeBDot() const {
+    // Get time delta between samples in seconds
+    double dt_seconds = this->m_magnetometer_sampling_period.count() / 1e6;
+    if (dt_seconds <= 0.0) {
+        return std::array<double, 3>{0.0, 0.0, 0.0};
+    }
+
+    // Compute BDot using 5-point Central Difference Formula
+    double x = (-1.0 * this->m_sampling_set[4].magnetic_field[0] + 8.0 * this->m_sampling_set[3].magnetic_field[0] +
+                -8.0 * this->m_sampling_set[1].magnetic_field[0] + 1.0 * this->m_sampling_set[0].magnetic_field[0]) /
```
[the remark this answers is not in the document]
nateinaction: It's true, I intentionally removed safeguards... I'll readd this one.

PR #170, opened 2025-12-02
Update test-integration to specify specific tests

hrfarmer:
You can now use the following syntax to specify specific test(s) from the integration test directory:
```bash
make test-integration imu_manager load_switch
```


ineskhou (approved):
LGTM! Runs great on my machine and passes tests

PR #186, opened 2025-12-07
Change ComQueue to 20

Lex-ari:


Mikefly123 (approved):
LGTM! 

This helps delay the explosions of the comQueue when we are emitting too much telemetry.

PR #216, opened 2025-12-18
Write new sequence number when dataIn_handler is called

hrfarmer:
# Write new sequence number when dataIn_handler is called

## Description

Updates the `dataIn_handler` in the `Authenticate` component to increment the sequence number on the board when it is called. I also set it to read the sequence number from the class instead of the file, although that can be reverted if necessary.

## How Has This Been Tested?

Ran on the board and the sequence number stays in sync with the ground station plugin

- [ ] Unit tests
- [x] Integration tests
- [ ] Z Tests

## Checklist


ineskhou (approved):
LGTM

PR #223, opened 2025-12-18
Command Loss Timer

ineskhou:
# Command Loss Timer

## Description

This adds a command loss timer to the authentication router. When the router has not received anything from the comm stack for COMM_LOSS_TIME seconds (defaulting to one day right now) it sends a message to go into safe mode. This will conserve power, but also execute a sequence from the to reset the radio (this is all in mode manafger)

Because only the Lora stack should be looked at for command loss timing, is it the only one actually hooked to the rate group and this the only one to command loss time

## How Has This Been Tested?

- [ ] Unit tests
- [ ] Integration tests
- [ ] Z Tests

Ran manually with many loss times (screenshots in the comments)

## Checklist

### DIFF
--- a/FprimeZephyrReference/Components/AuthenticationRouter/AuthenticationRouter.cpp
+++ b/FprimeZephyrReference/Components/AuthenticationRouter/AuthenticationRouter.cpp
@@ -16,6 +16,7 @@
 #include "Fw/Logger/Logger.hpp"
 #include "Os/File.hpp"
 #include "config/ApidEnumAc.hpp"
+#include <zephyr/drivers/rtc.h>
 
 constexpr const U8 OP_CODE_LENGTH = 4;  // F Prime opcodes are 32-bit (4 bytes)
 constexpr const U8 OP_CODE_START = 2;   // Opcode starts at byte offset 2 in the packet buffer


LeStarch (commented):
Revert to passive because this is the expectation of the router interface, add in a mutex to protect the file writes and memeber variable.

Boolean can be atomic.

ineskhou:
Screenshot 2025-12-24 at 1 29 46 AM

worked after 30, gonna run it for a few hours to make sure it works for longer periods

hrfarmer:
When I attached one of the ports myself though, something I noticed is that at least for me, I didn't see the command loss timer actually trigger until I did a reboot of the board by pressing the reset button.

ineskhou:
@hrfarmer
I didn't see the same issue on my end, but  I added the stuff we have for checking the timebase in the startup component, so if being in monotonic time was the issue, that would fix it. Just checking, when you ran it, did you change the COMM_LOSS_TIME to something observable bc it currently is set to one day. 
there is also the fact that if the comm_loss_start.bin exists already on the satelite, then you might have seen the message during the restart sequemce, and so you aren't seeing it until the second restart, you might have just started the GDS late? every restart that the command loss time is longer than the value it should be it send to go into safe mode rn, (this may be a design thing we want to change (just need to make the status written the filesystem) but thats a seperate thing that needs more convo about operation w/ s band radio).

hrfarmer (approved):
tested with a 60 seconds cooldown and i saw command loss triggering now, lgtm!

FprimeZephyrReference/Components/AuthenticationRouter/AuthenticationRouter.cpp
```diff
@@ -153,6 +168,69 @@ void AuthenticationRouter ::cmdResponseIn_handler(FwIndexType portNum,
...
+
+    Fw::ParamValid is_valid;
+    Fw::TimeIntervalValue command_loss_period = this->paramGet_COMM_LOSS_TIME(is_valid);
+    FW_ASSERT(is_valid == Fw::ParamValid::VALID || is_valid == Fw::ParamValid::DEFAULT);
+    Fw::Time command_loss_interval(command_loss_start.getTimeBase(), command_loss_period.get_seconds(),
+                                   command_loss_period.get_useconds());
+    Fw::Time command_loss_end = Fw::Time::add(command_loss_start, command_loss_interval);
+
+    Fw::Time current_time =
+        (command_loss_end.getTimeBase() == TimeBase::TB_PROC_TIME) ? this->get_uptime() : this->getTime();
```
nateinaction: What is the reason for doing this? If the time base is `TimeBase::TB_PROC_TIME`, the FPrime getTime() fn will return uptime.

PR #235, opened 2025-12-23
Fixes from 20225-12-22

LeStarch:

## Description

Fixes from 2025-12-22


Mikefly123 (approved):
LGTM! Saw everything working in the lab with file uplink via LoRa.

Fixes two issues:
1. The packet radio exploding because we are only able to sent payloads of up to 204bytes (rather than the full 252 bytes) due to the framing and authentication overhead
2. Refactors the CircuitPython LoRa passthrough to be way more user friendly and stable. Notably, there seemed to be some kinda race condition between the console and data USB UART connections that has now been solved

PR #247, opened 2025-12-31
B_dot 5-point central difference equation derivation write-up

evanjellison:
Made a derivation write-up for the 5-point central difference formula used for calculating B-dot.


nateinaction (changes requested):
Thank you for writing this up! I'm seeing some rendering issues in the markdown.

Image

PR #248, opened 2025-12-31
Revert clock tick change

nateinaction:
## Description
I included this in #79 before I saw @LeStarch's feedback on #245. Reverting.

I have verified detumble continues to work on the 50Hz rate group without this change.


Mikefly123 (approved):
LGTM!

PR #260, opened 2026-01-04
Comment out integration target

hrfarmer:
As of right now, I'm fairly certain something related to telemetry delay instantly freezes the board when I try to run integration tests. Given this issue + a ton of the tests not working properly at the moment anyways, I think we should just remove the target until the tests are in a state where they're actually useful.


ineskhou (approved):
LGTM! Saves us some time

PR #271, opened 2026-01-06
copy-secrets Makefile target

hrfarmer:
Adds `make copy-secrets` to make it easier to copy keys & authenticate file to your repo


ineskhou (approved):
makefile

PR #272, opened 2026-01-06
Update Startup Way

ineskhou:

## Description

We have identified 2 problems

1. The antenna deployer does save that it deployed on fail
2. the logic in which we were operating, mean that we only had about 3 seconds in integration the sat could be on. This was deemed unsafe as the sat could turn on during integration after last checkouts. The startup sequence was edited in order to guarentee a 45 min wait every time

## Related Issues/TIckets
- [ ] Unit tests
- [ ] Integration tests
- [ ] Z Tests

## Checklist

- [x] Have done a code review with


hrfarmer (approved):
YIPPEEE

PR #308, opened 2026-01-29
Update key makefile

ineskhou:
# Update key makefile
## Description

Adding copying local keys to make and changing the readme accordingly

## How Has This Been Tested?

- [ ] Unit tests
- [ ] Integration tests
- [ ] Z Tests

## Checklist

- [x] Have tested this PR on every supported board with correct board definitions

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ If you regenerate/replace the bootloader (or switch computers and flash a bootlo
 ```shell
 make copy-keys
 ```
-You also want to make sure the key the gds runs with is the same as the key on the board. For that, you want to make sure the key in FprimeZephyrReference/Components/Authenticate/AuthDefaultKey.h match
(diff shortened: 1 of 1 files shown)


hrfarmer (changes requested):
just for clarity that it's different from the signing key

Makefile
```diff
@@ -131,6 +131,10 @@ generate-auth-key: ## Generate AuthDefaultKey.h with a random HMAC key
 	fi
 	@echo "Generated $(AUTH_DEFAULT_KEY_HEADER)"
 
+.PHONY: copy-keys
+copy-keys:
```
nateinaction: If you remove the `PHONY` declaration and name the target `keys/proves.pem` it will only run if that file doesn't exist. This means that you can set this target to run automatically during the build step and remove the readme guidance.


```suggestion
keys/proves.pem:
```

PR #312, opened 2026-01-31
Automatically copy keys if they don't exist

nateinaction:
## Description

This PR automatically copies keys if they don't exist.


Makefile
hrfarmer: also needs to be changed here

hrfarmer (changes requested):

hrfarmer (approved):

PR #391, opened 2026-05-12
test(int): Add Stable Integration Tests Over the Radio

Mikefly123:
## Summary
Adds a second CI integration pass that communicates with the flight software over LoRa RF instead of the direct USB UART, validating that the radio link is functional end-to-end. Introduces the rf_unsafe pytest marker to gate tests that would sever the RF link, and makes numerous improvements to radio test reliability.

A bit of a journey to get here, but very cool now that we can implicitly validate that the radio works along with all of the other functions of the satellite. It does increase the time it takes for integration tests to run from around 10 minutes to 25-30 minutes though, so setting up a second CI runner that can allow for parallel runs would be really beneficial! 

## Changes
- RF Integration CI Pass
- Added integration-radio CI job that runs the full test suite via the v5d LoRa passthrough TTY (LORA_PASSTHROUGH_TTY)
integration-uart job now runs before integration-radio; added integration fan-in job to satisfy branch protection check
- CI bootstraps the sequence number over UART before starting the LoRa GDS to ensure command sequencing is correct
- CI restores clean hardware state after RF tests via Korad power-cycle and UART GDS restart
- Merged main's YAMCS round-trip CI test into the integration-uart job
- rf_unsafe Marker
- Added rf_unsafe pytest marker (defined in pytest.ini) to exclude tests that sever the RF link from the radio integration pass
- Marked tests as rf_unsafe: burnwire, antenna deployer, RTC alarm tests, reset manager, raw-bytes smoke test,test_00_setup_only transmit teardown
- test-integration-radio Makefile target filters with not flaky and not rf_unsafe
- test-integration (UART) filter reverted to not flaky only

## Radio Test Reliability Improvements
- Stabilization wait: 15-second wait after first TRANSMIT ENABLED + clear_histories() to drain boot-time event backlog before tests begin
- Fibonacci backoff with jitter: Retry delays follow the Fibonacci sequence (1, 1, 2, 3, 5, 8, 13 s) with ±50% random jitter to reduce half-duplex timing collisions
- Radio recovery: Automatically re-sends TRANSMIT ENABLED after 3 consecutive command failures to recover from dropped radio state
- RTC tests moved last: pytest_collection_modifyitems hook in conftest.py moves rtc_test.py to end of collection when --with-radio is active
- TMP112 retry loop: Outer retry loop wraps the full GetTemperature command + Temperature event assertion with 5 s event timeout
- Fail-fast: --exitfirst added to test-integration-radio so the run aborts on the first failure
- Command retries: --with-radio flag bumps retries to 5 per command via proves_send_and_assert_command

## Test File Cleanup
- Removed numeric prefixes: 0_radio_test.py and 1_lora_passthrough_test.py deleted; canonical names are radio_test.py and lora_passthrough_test.py
- Removed RADIO_DEBUG_PLAN.md
- Removed requires_hw_watchdog marker from watchdog_test.py
- Removed pytestmark skip from drv2605_test.py

## Telemetry & Observability
- Added telemetry sampler fixture and CI artifact upload for integration test runs

## How Has This Been Tested?
-[x] Integration tests (UART pass in CI)
-[x] Integration tests (RF/LoRa pass in CI)


Mikefly123:
@copilot resolve the merge conflicts in this pull request

PROVESFlightControllerReference/project/config/LoRaCfg.hpp
```diff
@@ -6,7 +6,7 @@
 namespace LoRaConfig {
 const U32 FREQUENCY = 437400000;               //!< LoRa frequency in Hz
 lora_signal_bandwidth BANDWIDTH = BW_125_KHZ;  //!< LoRa bandwidth
-const I8 TX_POWER = 23;                        //!< LoRa transmission power in dBm
```
nateinaction, line 9: Keep at `23`

PROVESFlightControllerReference/ReferenceDeployment/Main.cpp
```diff
@@ -21,7 +21,7 @@
 const struct device* ina219Sys = DEVICE_DT_GET(DT_NODELABEL(ina219_0));
 const struct device* ina219Sol = DEVICE_DT_GET(DT_NODELABEL(ina219_1));
 const struct device* serial = DEVICE_DT_GET(DT_NODELABEL(cdc_acm_uart0));
-const struct device* lora = DEVICE_DT_GET(DT_NODELABEL(lora0));
```
nateinaction, line 24: Keep `DEVICE_DT_GET`

PROVESFlightControllerReference/test/int/0_radio_test.py
nateinaction, line 1: Remove this file, it now lives at `radio_test.py`

PROVESFlightControllerReference/test/int/1_lora_passthrough_test.py
nateinaction, line 1: rename to just `lora_passthrough_test.py`

PROVESFlightControllerReference/test/int/drv2605_test.py
```diff
@@ -12,6 +12,8 @@
 from fprime_gds.common.models.serialize.time_type import TimeType
 from fprime_gds.common.testing_fw.api import IntegrationTestAPI
 
+pytestmark = pytest.mark.skip(reason="needs in-lab review")
+
```
nateinaction: remove

PROVESFlightControllerReference/test/int/watchdog_test.py
```diff
@@ -87,6 +89,7 @@ def test_02_system_stays_running_with_watchdog(
 
 
 @pytest.mark.slow
+@pytest.mark.requires_hw_watchdog
```
nateinaction: remove

Makefile
```diff
@@ -227,7 +227,7 @@ test-unit: ## Run unit tests
 test-integration: uv ## Run integration tests (set TEST=<name|file.py> or pass test targets)
 	@DEPLOY="build-artifacts/zephyr/fprime-zephyr-deployment"; \
 	TARGETS=""; \
-	FILTER="not flaky"; \
+	FILTER="not flaky and not requires_hw_watchdog"; \
```
nateinaction: revert this to just `FILTER="not flaky";`

Makefile
```diff
@@ -248,6 +248,18 @@ test-integration: uv ## Run integration tests (set TEST=<name|file.py> or pass t
...
 
+.PHONY: reenable-lora
+reenable-lora: uv ## Re-enable LoRa transmit before the RF integration pass
+	$(UV_RUN) pytest PROVESFlightControllerReference/test/int/0_radio_test.py::test_00_setup_only \
+	    --deployment build-artifacts/zephyr/fprime-zephyr-deployment
+
+.PHONY: test-integration-radio
+test-integration-radio: uv ## Run integration tests via LoRa passthrough GDS
+	@DEPLOY="build-artifacts/zephyr/fprime-zephyr-deployment"; \
+	FILTER="not flaky and not requires_hw_watchdog and not rf_unsafe"; \
```
nateinaction:
```suggestion
	FILTER="not flaky and not rf_unsafe"; \
```

pytest.ini
```diff
@@ -2,3 +2,4 @@
 markers =
     slow: marks tests as slow
 	flaky: marks tests as flaky
+	requires_hw_watchdog: needs HW watchdog wired/enabled (skipped on v5e for now)
```
nateinaction: remove

.github/workflows/ci.yaml
```diff
@@ -100,6 +100,7 @@ jobs:
             mcuboot.elf
             bootable.uf2
             bootable.signed.hex
+            build-fprime-automatic-zephyr/zephyr/zephyr.elf
```
nateinaction: I don't see a need for this. Let's remove.

RADIO_DEBUG_PLAN.md
nateinaction, line 1: remove file

Mikefly123:
@copilot implement all of the changes that were requested by @nateinaction

Mikefly123:
@copilot Implement the last two suggestions from the CodeRabbit Review

Mikefly123:
@copilot there is a bug in the behavior of the radio integration tests:
```py
        if request.node.name != "test_00_setup_only":
>           proves_send_and_assert_command(
                fprime_test_api,
                f"{lora}.TRANSMIT",
                ["DISABLED"],
            )

PROVESFlightControllerReference/test/int/radio_test.py:31: 
```
We cannot assert on the Transmit Disable command if we are only monitoring the satellite via the radio link as sending this command will stop any further events from leaving the system via the radio. This includes the event that confirms this command was received. 

Ensure that any command that we expect to be confirmed in over the radio integration tests can only take place when the radio transmit is enabled. If the transmit is disabled but we still want to assert an event has occurred we will need to connect to the USB debug connection on the integration hardware

Mikefly123:
@copilot address the comments left by CodeRabbit in its review. 

Also we do want to ensure that the integration test hardware returns to a clean state after the radio tests are done running. To do that we want to ensure the transmit is disabled. This could be accomplished explicitly by command or we could also do a hard on / off power cycle with the Korad Control to clear any residual state

Mikefly123:
@copilot the TMP112 test seems to be flaky but often will pass on retries. Can you modify the test code to allow for more retries on the TML112 test before failing?

Mikefly123:
@copilot let's implement fail fast logic to the radio integration tests. Right now the radio integration tests will continue running for a very long time even if there is no chance of continued success. See the last CI run for an example of this

Mikefly123:
@copilot let's skip the burnwire test in over the radio integration testing right now due to a need to refactor that test for better over the air reliability

PROVESFlightControllerReference/test/int/radio_test.py
```diff
@@ -20,17 +20,25 @@
...
-    yield
+
+
+def test_00_setup_only(fprime_test_api: IntegrationTestAPI, start_gds):
+    """Enable LoRa transmit; used as a standalone re-enable step before the RF integration pass."""
     proves_send_and_assert_command(
         fprime_test_api,
         f"{lora}.TRANSMIT",
-        ["DISABLED"],
+        ["ENABLED"],
```
nateinaction: We don't want to bleed state over multiple tests like this. Instead add this test to rf_unsafe

Mikefly123:
https://github.com/Open-Source-Space-Foundation/proves-core-reference/issues/402

Mikefly123:
@copilot for our radio based integration tests, the first time we enable transmit on the radio there may be a large queue of event messages that need to be cleared before we are able to receive events resulting from the commands we send in integration tests. Let's add a stabilization wait after the first time we transmit enable to allow these messages to empty out, before we go ahead and send new commands to the system. The events generated by new commands might be getting dropped by a comm queue overflow

Mikefly123:
@copilot address the merge conflicts from main in this branch. We want to merge in main and keep all of the new YAMCS CI testing as part of the UART Integration Tests.

Mikefly123:
@copilot check if the test fixtures that send commands in the TMP112 test have the same retry logic as the other send and asserts. Not having retries may be the reason we are seeing so much flakiness on that test

Mikefly123:
@copilot I think there is a possibility that the radio test flake is due to accidental timing collisions between the FSW transmitting / receiving. Our communications are only half duplex, so if we try to send (uplink) a command to the FSW via the radio at the same time the radio is being blocked by sending events or telemetry to us (downlink). Perhaps let's try randomizing the time interval between retries or using a Fibonacci type backoff to get around this issue.

Mikefly123:
@copilot Let's try moving the RTC tests in the Radio Integration tests to last. I think the changing of the time that happens as a part of the RTC test may be causing instability in the test setup for the test that occurs immediately afterwards (resulting in the flakey TMP112 test)

Mikefly123:
@copilot after our recent changes, the last IMU Manager test is consistently flaking now. Interestingly, it is now the only test file that has three tests in sequence with each other, previously (before implementing skips) the RTC test would also start flaking on the third test in the file. 

Could this potentially be an issue with how we are starting up and tearing down the radio transmit enable state after every test? Doing so sequentially without enough of a break (say during a string of successful tests) may cause packets to get dropped, resulting in tests flaking due to a bad radio state that is not reset until the next test attempt. 

Consider modifying the behavior of the radio integration test retries to also retry sending the Transmit Enable command after three subsequent failures to get a response for a command

Mikefly123:
@copilot Just two more things to clean up! Address the `lint` failure by adding the required docstrings to appease Interrogate. Also look into why there is still a pending "integration" check when no actions are scheduled to run for it

Mikefly123:
@copilot Update the PR message to reflect the work done in this branch

ineskhou (dismissed):
Great Documentation, especially of issues and stuff, and seems to be passing well!

LGTM!

nateinaction (approved):

PR #395, opened 2026-05-16
Authenticate Refactor: Phase 1

nateinaction:
## Description

Hey team,

I'm working to reduce the complexity of and genericize our authentication stack. This is the first in a series of PRs to accomplish that.

This PR:
- ~Does not change any functionality! 😄~
  - Changes behavior of command loss timer to use a rate group tick instead of depending on the file system or the RTC
    - Solves #404
    - Solves #424
- Move mode changing from `AuthenticateRouter` to `ModeManager`
- Renames `AuthenticateRouter` to `ProvesRouter` since its function doesn't really touch authentication and the router it's named after is called `FPrimeRouter`
- Removed unused events and telemetry
- Adds a test for command loss behavior

Additional thoughts:
- `ModeManager` is pretty complex with no unit tests. We should refactor soon with a focus on verifying behavior with unit testing.

Where we are now:

```mermaid
flowchart TD
    TcDeframer <--> Authenticate
    Authenticate <--> SpacePacketDeframer
    SpacePacketDeframer <--> AuthenticateRouter
```

Where we're going:

```mermaid
flowchart TD
    TcDeframer <--> TcSecurityDeframer
    TcSecurityDeframer <--> SecurityValidator
    TcSecurityDeframer <--> SpacePacketDeframer
    SpacePacketDeframer <--> ProvesRouter
```

## How Has This Been Tested?

- [ ] Unit tests
- [x] Integration tests
- [ ] Z Tests

Verified commands working:

Captura de pantalla 2026-05-16 a la(s) 15 14 10


PROVESFlightControllerReference/Components/ModeManager/docs/sdd.md
```diff
@@ -81,13 +86,14 @@ classDiagram
...
 
-Voltage thresholds are configurable via F-Prime parameters:
+Voltage thresholds and command loss timeout are configurable via F-Prime parameters:
 
 | Parameter | Type | Default | Description |
 |---|---|---|---|
 | SafeModeEntryVoltage | F32 | 6.7 | Voltage (V) below which safe mode is entered |
 | SafeModeRecoveryVoltage | F32 | 8.0 | Voltage (V) above which safe mode can be exited |
 | SafeModeDebounceSeconds | U32 | 10 | Consecutive seconds required for transitions |
+| COMM_LOSS_TIME | Fw.TimeIntervalValue | {seconds=86400} | Time without an authenticated packet before command loss safe mode entry (default: 3 days) |
```
Mikefly123: `86400` seconds is only 1 day. 
```suggestion
| COMM_LOSS_TIME | Fw.TimeIntervalValue | {seconds=3*60*60*24} | Time without an authenticated packet before command loss safe mode entry (default: 3 days) |
```

docs-site/components/ProvesRouter.md
```diff
@@ -0,0 +1,59 @@
...
+| `output` | `unknownDataOut` | `Svc.ComDataWithContext` | Port forwarding unknown data (useful for adding custom routing rules with a project-defined router) |
+| `output` | `bufferAllocate` | `Fw.BufferGet` | Port for allocating buffers, allowing copy of received data |
+| `output` | `bufferDeallocate` | `Fw.BufferSend` | Port for deallocating buffers |
+| `output` | `packetRouted` | `Fw.Signal` | Emitted after each received packet is processed; used to reset command loss timer in ModeManager |
+
+## Requirements
+
+| Name | Description | Rationale | Validation |
+| ---- | ----------- | --------- | ---------- |
+SVC-ROUTER-001 | `Svc::ProvesRouter` shall route packets based on their packet type as read from the `ComCfg::FrameContext` APID field (`context.get_apid()`) | Routing mechanism of the F´ comms protocol | Unit test |
```
Mikefly123: Unit tests are discussed as Validation methods here, but we haven't implemented any of them yet. Should we capture a TODO or a ticket to add these later?

Mikefly123 (commented):
Some quick comments! More review to come soon. It generally is looking good to me though

PROVESFlightControllerReference/Components/ProvesRouter/ProvesRouter.cpp
```diff
@@ -0,0 +1,142 @@
...
+    // and ProvesRouter can handle the deallocation of the unknown buffer when it returns on bufferReturnIn
+    Fw::Buffer copy = this->allocateCopy(packetBuffer, ProvesRouter_AllocationReason::USER_BUFFER);
+    if (copy.isValid()) {
+        // Send the copied buffer to connected components
+        this->unknownDataOut_out(0, copy, context);
+
+        // Notify connected components that a packet was routed
+        this->notifyPacketRouted();
+    }
+}
```
Mikefly123, line 104: I think this is okay for an initial implementation (because it matches default F Prime Behavior), but we may want to reconsider the `notifyPacketRouted` call here because unknown packets just get dropped silently by the router. 

I think command loss behavior is the main focus of this PR and in that regard I don't think unknown packets should count as having command control of the satellite since no action is actually taken when an unknown packet is received.

PROVESFlightControllerReference/test/int/mode_manager_test.py
Mikefly123, line 1: Rather than using an RTC time set for this test, what if we just set the command loss timer parameter to something very small, like `5s`? That is perhaps less fight like, but feels safer since it doesn't mess with the RTC, which is something that we cannot guarantee gets cleaned up correctly between test failures (unlike a soft parameter which will reset to default on reboot). 

I don't think this is a blocking issue if you want to just go for it, just something to consider.

PROVESFlightControllerReference/Components/ModeManager/ModeManager.cpp
```diff
@@ -506,4 +520,37 @@ F32 ModeManager ::getCurrentVoltage(bool& valid) {
...
+    Fw::ParamValid paramValid;
+    Fw::TimeIntervalValue commLossPeriod = this->paramGet_COMM_LOSS_TIME(paramValid);
+    FW_ASSERT(paramValid == Fw::ParamValid::VALID || paramValid == Fw::ParamValid::DEFAULT);
+
+    // Check if current time has exceeded the command loss start time + command loss period
+    Fw::Time commLossInterval(this->m_lastPacketRoutedTime.getTimeBase(), commLossPeriod.get_seconds(),
+                              commLossPeriod.get_useconds());
+    Fw::Time commLossEnd = Fw::Time::add(this->m_lastPacketRoutedTime, commLossInterval);
+    Fw::Time currentTime = this->getTime();
+    if (currentTime > commLossEnd) {
```
Mikefly123: This is potentially not safe enough against timebase changes. We should guard against changes or only use monotonic time, which is always available as long as the microcontroller is running. We are storing `this->m_lastPacketRoutedTime` in RAM anyways, so there is not currently a benefit to using real time (other than accuracy). 

Review Comment Generated by Copilot: 
> Potential bug: `commandLossCheck()` assumes `getTime()` always returns the same time base as `m_lastPacketRoutedTime`. If a packet is routed when `getTime()` is using one base (for example real time), and later `commandLossCheck()` runs after the clock source has switched (for example to monotonic), then `currentTime` and `commLossEnd` are no longer comparable in a meaningful way. That can cause a false `COMMAND_LOSS` trigger, a missed timeout, or an assert depending on `Fw::Time` comparison semantics. Since this is timeout logic, it would be safer to use a monotonic source only, or at least guard against `currentTime.getTimeBase() != m_lastPacketRoutedTime.getTimeBase()` before doing the add/compare.

Mikefly123 (changes requested):
Hey @nateinaction! Great work on this so far. After further review, I think my comment on hardening against timebase changes is the only blocking issue I have with this Pull Request. Otherwise looking forward to getting it in!

PROVESFlightControllerReference/Components/ProvesRouter/ProvesRouter.cpp
nateinaction: Maybe not for this PR but I'm curious... what would you want the behavior of unknown packets to be?

PROVESFlightControllerReference/Components/ModeManager/ModeManager.cpp
nateinaction: @collinc04 is going to take this on in #404 / #424. Are you open to a fast follow with those changes? We have a plan to make command-loss based on a counter driven by a rate group instead of by the RTC.

PROVESFlightControllerReference/test/int/mode_manager_test.py
nateinaction: @collinc04 will need to rework this test for his changes as well.

PROVESFlightControllerReference/Components/ModeManager/ModeManager.cpp
nateinaction: nvm. We're going to merge in the counter driven command loss timer here.

PROVESFlightControllerReference/Components/ModeManager/ModeManager.cpp
```diff
@@ -506,4 +523,27 @@ F32 ModeManager ::getCurrentVoltage(bool& valid) {
...
+void ModeManager::commandLossCheck() {
+    // Protect against concurrent access to command loss state
+    Os::ScopeLock lock(this->m_commandLossMutex);
+
+    // Get command loss period from parameter
+    Fw::ParamValid paramValid;
+    Fw::TimeIntervalValue commLossPeriod = this->paramGet_COMM_LOSS_TIME(paramValid);
+    FW_ASSERT(paramValid == Fw::ParamValid::VALID || paramValid == Fw::ParamValid::DEFAULT);
+
+    if (this->m_commandLossCounter >= commLossPeriod.get_seconds() && this->m_mode == SystemMode::NORMAL) {
```
Mikefly123: I think this is not the correct behavior. If I'm reading this line correctly if we are already in SAFE mode we will bypass the command loss check and never have the command loss timer fire. 

This would potentially create a situation where the command loss timer never fires and the satellite has no automatic recovery if (for example) an operator commanded the satellite to a bad (unknown) radio state AND manually set a SAFE mode before doing so. 

Unless I am mistaken on how it is used here, we should drop the mode check on this statement and replace it the debounce like in the original AuthenticationRouter.

Mikefly123 (changes requested):
Hey @nateinaction! Thanks for nudging this along. I think there is one last behavioral tweak we should make before moving forward with merging this in.

Mikefly123 (approved):
LGTM! Thanks for the diligence on getting this through

PR #398, opened 2026-05-18
F Prime v4.2.2 upgrade

nateinaction:
## Description

This PR upgrades F Prime to v4.2.2

### CI: enforce that the Zephyr console stays disabled

The F' downlink and the Zephyr console share the same USB CDC ACM device
(`cdc_acm_uart0`). When the console is enabled, `printk` / `Os::Console` text
is interleaved with the binary CCSDS TM frames on that UART, which desyncs the
GDS deframer so no telemetry or command acks can be parsed. `prj.conf` now
disables the console (`CONFIG_CONSOLE` / `CONFIG_UART_CONSOLE`) so the downlink
UART carries frames only.

To prevent a regression from reaching `main`, this adds a CI guard:

- `scripts/check_console_disabled.py` — fails if `CONFIG_CONSOLE=y` or
  `CONFIG_UART_CONSOLE=y`. It inspects the generated Zephyr `.config` (the
  compiled state), so it also catches the console being re-enabled via a board
  defconfig or a Kconfig default, not just a direct `prj.conf` edit.
- `make check-console-disabled` — the same check as a Make target for local use.
- A new **"Ensure console disabled"** step in the build job runs it right after
  `make build`, so the build (a required check) fails if the console is on.

### Integration test markers for add-on hardware

Several integration tests only pass when hardware beyond the flight controller
board is attached. On a bare board they fail for environmental reasons, which
made it hard to tell a real regression from "that board just isn't plugged in."

Following the existing `uart_only` convention (mark the exceptional tests; leave
everything else unmarked), this adds markers naming the hardware each test needs:

| Marker | Requires | Tests |
| --- | --- | --- |
| `requires_face` | a face board (TMP112 / VEML6031 / DRV2605) | drv2605, tmp112, veml6031 |
| `requires_antenna` | the antenna board + burnwire capacitor | burnwire |
| `requires_battery` | the battery board with power at the terminals | power_monitor, burnwire |
| `requires_watchdog_jumper` | the JP6 watchdog jumper bridged | watchdog `test_03` |

How this helps testing:

- **Bench runs on a bare board are one flag.** `conftest.py` adds
  `--bare-flight-controller-board`, which skips any test carrying those markers with
  a reason naming the missing hardware, instead of a long `-m "not ... and not ..."`
  expression:
  ```
  make test-integration PYTEST_ARGS=--bare-flight-controller-board
  ```
- **Failures are self-explanatory.** A skip now reads e.g. `requires the JP6
  watchdog jumper bridged`, so it is obvious *why* a test did not run rather than
  looking like a flaky failure.
- **Composes across rigs.** A partially-populated rig can exclude only what it
  lacks (e.g. `-m "not requires_antenna and not requires_battery"`) and still run
  everything else.
- **Safe default for new tests.** Unmarked tests run on a bare board; a test is
  only tagged when it genuinely needs add-on hardware.


PROVESFlightControllerReference/project/config/ComCfg.fpp
```diff
@@ -12,7 +12,10 @@ module ComCfg {
     dictionary constant SpacecraftId = 0x0044
 
     @ Fixed size of CCSDS TM frames
-    dictionary constant TmFrameFixedSize = 248  # Needs to be at least COM_BUFFER_MAX_SIZE + (2 * SpacePacketHeaderSize) + 1
```
nateinaction, line 15: I think this change is not correct. Isn't this a radio limitation? I remember something about this when working on the CircuitPython stuff. Need radio integration tests.

Mikefly123: Yes this is a limitation of the max packet size of the radio. If I recall correctly the radio module has a 4 byte header and then the Zephyr Radio driver has another 4 byte header leaving us with 248 (256 - 4 -4 = 248) to use for F Prime.

Mikefly123:
CI is green — root cause of the YAMCS round-trip failure found and fixed.

**Root cause** (`2d35b17`): fprime-yamcs 0.1.3's `get_dictionary_constants()` returns constants in *dictionary-file* order, while `construct_temporary_configuration()` indexes the result positionally (`constants[0]` = `TmFrameFixedSize`, `constants[1]` = `SpacecraftId`). F´ v4.2.2 dictionaries list `ComCfg.SpacecraftId` *before* `ComCfg.TmFrameFixedSize` (v4.1.1 was the reverse), so YAMCS's `UdpTmFrameLink` was silently configured with `frameLength=67, spacecraftId=248`. Every TM frame was rejected as invalid — archive diagnostics showed `39 received / 0 valid / 39 invalid` — so no packets, parameters, or events ever reached YAMCS and the `NoOpReceived` wait timed out. The TC uplink was never at fault (the adapter rebuilds TC frames itself from `SPACECRAFT_ID` env, and its frame bytes are provably identical to the GDS chain's).

**Fix**: `tools/apply-yamcs-constants-order-fix.py`, applied during `make fprime-venv` like the existing fprime-yamcs fixers, makes the lookup order-deterministic and fails the build loudly if the upstream source changes. This should also be reported upstream to fprime-yamcs.

Also added in this push, worth keeping: a board power-cycle between the GDS and YAMCS phases (makes the phases state-independent), and a "Dump YAMCS archive diagnostics on failure" step (packets / `CommandsDispatched` samples / link stats / events) — the link stats are what cracked this.

Note for the runner: one earlier artifact (run 29725434461) was built with a stale `fprime-zephyr` checkout (`1df87d4` instead of the pinned `17cd7f3` — visible in the dict's `metadata.libraryVersions`). Harmless here, but the build runner's submodule state is worth keeping an eye on.

https://claude.ai/code/session_01SmGJ3xFxWRYWDzJyfCCwUF

Mikefly123 (approved):
LGTM!

PR #399, opened 2026-05-21
fix(StartupManager): reject corrupt quiescence file to break boot-loop

Mikefly123:

## Description
A partial flash write to /quiescence_start.bin (e.g. power loss mid-write) can leave a Fw::Time on disk whose useconds field is outside [0, 999999]. On boot, update_quiescence_start reads it back and passes it to the static Fw::Time::add at StartupManager.cpp:189, which trips FW_ASSERT in Fw/Time/Time.cpp:141 (uSeconds < 1999999) followed by line 43 (useconds < 1000000) — observed on hardware with useconds = 0x00FFFFFF (16,777,215), the signature of unprogrammed NOR flash bytes. Watchdog reboot loop, no recovery without external reflash, and the corrupt file persists across firmware versions so any branch hits the same fault.

Validate useconds after read; treat out-of-contract values as a read failure so the existing init-from-getTime + rewrite path takes over.


ineskhou (approved):
LGTM! Cool solution!!!

PR #412, opened 2026-06-03
Remove uart dependency on radio tests

nateinaction:
This PR removes the uart dependency on the radio tests.

Benefits:
- We can rerun the radio tests even when the uart tests are failing
- If we ever have a second integration test harness, we can run both tests at the same time

This PR also removes the `integration` step since it doesn't appear to be necessary right now.


Mikefly123 (approved):
LGTM!

PR #413, opened 2026-06-05
chore(deps): bump fprime-yamcs to 0.1.3 (and fprime-xtce to 0.1.2)

yudataguy:
## What

Bumps the [`fprime-yamcs-reference`](https://github.com/Open-Source-Space-Foundation/fprime-yamcs-reference) dependency to pick up its uplink/downlink work and the `My*` → `Fprime*` class renames.

- **`requirements.txt`**: `fprime-yamcs` `0.1.2` → `0.1.3`, `fprime-xtce` `0.1.1` → `0.1.2` (matches the upstream reference pins).
- **`tools/apply-yamcs-instance-config-fix.py`**: also match the renamed `Fprime{Packet,Command}Pre/Postprocessor` classes.

## Why the patch-script change matters

The 0.1.3 template renamed its placeholder classes (`MyPacketPreprocessor` → `FprimePacketPreprocessor`, `MyCommandPostprocessor` → `FprimeCommandPostprocessor`) and added a new `FprimeFilePacketService` (file-downlink reassembly — the "downlink" piece).

Our `apply-yamcs-instance-config-fix.py` does exact-string `.replace()` on those class names. Against 0.1.3 those two replacements would silently no-op — and because the script's *other* edits (frame length, ProcessRunner removal, rootContainer) still fire, it would **not** hit its "no changes made" guard. It would quietly leave custom Java class names in the generated config. Both old and new spellings are now handled.

## Compatibility verification (against the real 0.1.3 wheel)

- ✅ `__main__.py` is **byte-identical** to 0.1.2 → `patches/fprime-yamcs-noapp-path.patch` still applies.
- ✅ `events/processor.py` still contains the `while True: pass` busy-wait → `apply-events-cpu-fix.py` still applies.
- ✅ Hardened `apply-yamcs-instance-config-fix.py` run against the 0.1.3 template produces `CfsPacketPreprocessor` + 248-byte frame/packet lengths + `rootContainer`, with no leftover custom class refs.

## Runtime impact

None expected. The runtime YAMCS config is the vendored `yamcs/yamcs-data/etc/` (passed via `--yamcs-config-dir`), which already uses the stock `org.yamcs.tctm.cfs.CfsPacketPreprocessor` over UDP CCSDS framing — insulated from the rename. The new `FprimeFilePacketService` lives only in the package template and is **not** wired into the vendored config (out of scope for this bump).


Mikefly123 (approved):
LGTM! 

Finally merging this now that the CI is back online

PR #414, opened 2026-06-21
test(radio): CONTINUOUS_WAVE repeated-call regression test (#207)

### LINKED ISSUE #207: [BUG] Using LoRa Continuous Wave Function Fails After First Try
### Short summary

Calling the LoRa Continuous Wave function causes the component to fail.

### Steps to reproduce

1. Flash `main`
2. Send a continuous wave command
3. Observe failure

### Expected behavior

Sending the continuous wave commands should succeed when called multiple times.

### Actual behavior

The continuous wave function throws an execution error on every call.

Image

The first time it is called though, you can verify that it does successfully transmit exactly once when observed with an SDR. Attempts to start the radio again fail unless you reset.

### Logs / crash output

_No response_

### Target board

RP2350 (Pico 2)

Mikefly123:
## What

Adds `test_02_continuous_wave_repeated` to `PROVESFlightControllerReference/test/int/radio_test.py`, a regression test for #207 (*"Using LoRa Continuous Wave Function Fails After First Try"*).

The test:
- sends `lora.CONTINUOUS_WAVE` **twice** (waiting out the burst between calls) and asserts **both** commands return OK,
- asserts no `ConfigurationFailed` event is logged by the second call,
- asserts a normal `TRANSMIT` path still works afterwards (i.e. the modem is not wedged).

Marked `uart_only`, consistent with the existing radio integration test.

## Depends on

The firmware fix in Open-Source-Space-Foundation/fprime-zephyr#11. Before that fix the second `CONTINUOUS_WAVE` returned `EXECUTION_ERROR` and the modem stayed wedged (no TX/RX) until a board reset — so this test fails on current firmware and passes once the submodule is bumped to include the fix.

## Verification

The fix + this scenario were verified on GRC hardware (RP2350 / SX127x) with a GDB watchpoint on `dev_data.modem_usage`: across two consecutive `CONTINUOUS_WAVE` commands the modem returns to FREE via the radio's TxTimeout handler and is re-armed to BUSY for async RX after each call; both calls physically transmit and the FSW stays responsive.

Closes #207


ineskhou (approved):
Updates tests passes tests LGTM!

PR #415, opened 2026-06-30
feat(yamcs): promiscuous TM adapter — accept any spacecraft ID

Mikefly123:
## Summary

- Replaces the sync-header-based frame scan (which filtered by SCID) with a CRC-16/CCITT-only HUNT/LOCK state machine, so the adapter accepts TM frames from any satellite regardless of the embedded spacecraft ID
- Before forwarding to YAMCS, frames are SCID-normalized to `--spacecraft-id` (default 68) and the CRC is recomputed — YAMCS's `spacecraftId` filter still passes with no config change
- Logs the real SCID on first lock (`[TM] acquired lock: SCID=69 → normalizing to 68`) for visibility

## Motivation

A single GRC board feeds a single YAMCS instance and may receive telemetry from multiple PROVES satellites whose spacecraft IDs are not yet known. This removes the requirement to know the SCID before running the adapter.

## Test plan

- [ ] `make yamcs UART_DEVICE=/dev/cu.<GRC>` launches cleanly against a single satellite (SCID=68, no normalization path exercised)
- [ ] Connect a second satellite with a different SCID and confirm `acquired lock` log line appears and telemetry flows into YAMCS

### DIFF
--- a/.claude/skills/radio-link-testing/OTA-BEST-PRACTICES.md
+++ b/.claude/skills/radio-link-testing/OTA-BEST-PRACTICES.md
@@ -1,0 +1,164 @@
+# OTA uplink best practices
+
+Operational checklist for ground-to-flight firmware/file delivery over the LoRa
+link. Detailed mechanics are in the companion docs; this is the operator's quick
+reference.
+
+See also:
+- [REFERENCE.md](REFERENCE.md) — rig topology, ports, auth, GDS launch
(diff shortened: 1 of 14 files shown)


Mikefly123:
Closing — branch carried unrelated radio-link-testing commits. Replacing with a focused PR containing only the YAMCS adapter changes.

PR #416, opened 2026-06-30
feat(yamcs): promiscuous TM adapter — accept any spacecraft ID

Mikefly123:
## Summary

- Replaces the sync-header accept-list scan in `_forward_tm_serial` with a CRC-16/CCITT-only HUNT/LOCK state machine — no upfront SCID enumeration required
- Frames from any spacecraft ID are accepted; before forwarding to YAMCS, the SCID/VCID bytes are normalized to `--spacecraft-id` and the CRC is recomputed so YAMCS's `spacecraftId` filter always passes
- Per-SCID frame counters are retained in stats output so operators can see which satellites were heard

## Motivation

A single GRC board acts as the shared ground radio for multiple PROVES satellites whose spacecraft IDs are not yet known. Previously the adapter required listing every expected SCID via `--spacecraft-id`; now it captures telemetry from any satellite that comes overhead.

## Test plan

- [ ] `make yamcs UART_DEVICE=/dev/cu.<GRC>` launches and adapter logs `Promiscuous: accepting any SCID, normalizing to SCID=68`
- [ ] A satellite with SCID=68 flows through without normalization
- [ ] A satellite with a different SCID logs `acquired lock: SCID=<X> → normalizing to 68` and telemetry appears in YAMCS

### DIFF
--- a/tools/yamcs/proves_adapter.py
+++ b/tools/yamcs/proves_adapter.py
@@ -5,6 +5,12 @@ Routes TM frames from the spacecraft to YAMCS (via UDP) and TC frames from
 YAMCS (via UDP) back to the spacecraft after wrapping them with the HMAC-SHA256
 authentication header/trailer required by the FSW Authenticate component.
 
+The TM side is promiscuous: it accepts frames from any CCSDS spacecraft ID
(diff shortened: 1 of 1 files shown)


Mikefly123:
Closing in favor of making #418 work!

PR #433, opened 2026-07-11
Fix test issues

nateinaction:
## Summary
- Disable `CONFIG_CONSOLE`/`CONFIG_UART_CONSOLE`/`CONFIG_PRINTK` in `prj.conf`. The Zephyr console shares `cdc_acm_uart0` with the F' downlink; leaving it enabled interleaves printk/`Os::Console` text with the binary CCSDS TM frames and desyncs the GDS deframer, breaking telemetry and command acks.
- Add `scripts/check_console_disabled.py` + `make check-console-disabled`, run in the build job's CI, to fail the build if the *compiled* Zephyr `.config` shows the console re-enabled — catches regressions from a board defconfig or Kconfig default, not just direct `prj.conf` edits.
- Add pytest markers (`requires_face`, `requires_antenna`, `requires_battery`, `requires_watchdog_jumper`) for integration tests that need add-on hardware, plus a `--bare-flight-controller-board` flag to skip them, so the suite can run more easily on a bare flight controller board without a long `-m` expression:
  ```
  make test-integration PYTEST_ARGS=--bare-flight-controller-board
  ```
- Also includes a prior fix: use `timedelta` for the `pytest.approx` abs tolerance in `rtc_test`.

## Test plan
- [x] CI build job passes `check-console-disabled`
- [ ] UART integration tests run without the command/event dropouts seen previously

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -92,6 +92,13 @@ jobs:
         run: |
           make build
 
+      - name: Ensure console disabled
+        # The Zephyr console shares cdc_acm_uart0 with the F' downlink; if it is
+        # re-enabled, console text interleaves with the CCSDS TM frames and
+        # desyncs the GDS deframer. Fail the build so this cannot reach main.
+        run: |
+          make check-console-disabled
+
       - name: Upload build artifacts
         uses: actions/upload-artifact@v4
         with:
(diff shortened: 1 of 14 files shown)


Mikefly123 (approved):
LGTM!

PR #407, opened 2026-05-25
Rtc test fix

collinc04:
Functionality remains the same but it looks like it is more common convention to use assert instead

# Pull Request Title (e.g., Feature: Add user authentication)

## How Has This Been Tested?

- [ ] Unit tests
- [ ] Integration tests
- [ ] Z Tests
- [ ] Manual testing (describe steps)

## Checklist

- [ ] Written detailed sdd with requirements, channels, ports, commands, telemetry defined and correctly formatted and spelled
- [ ] Have written relevant integration tests and have documented them in the sdd
- [ ] Have done a code review with
- [ ] Have tested this PR on every supported board with correct board definitions

PR #435, opened 2026-07-11
fix(test): rtc test_01 TypeError under pytest 9.1 (datetime approx)

Mikefly123:
## Problem

Integration tests on branches cut from `main` fail intermittently in `test_01_time_set`:

```
FAILED PROVESFlightControllerReference/test/int/rtc_test.py::test_01_time_set
TypeError: absolute tolerance for datetime/timedelta must be a timedelta, got int
```

(e.g. [run 29165181551](https://github.com/Open-Source-Space-Foundation/proves-core-reference/actions/runs/29165181551) on `feat/psram-v5d`.)

## Root cause

`test_01_time_set` compares **datetimes** with `pytest.approx(..., abs=30)` — an **int** tolerance. pytest 9.1 added first-class datetime support to `approx` and now requires the tolerance to be a `timedelta`, raising `TypeError` otherwise. CI installs pytest unpinned, so runs that resolve pytest ≥ 9.1 fail while runs that resolve 8.3.3 pass — which is why the breakage appears branch-by-branch while `main`'s last run (pre-drift) is green.

On old pytest these lines were silently meaningless anyway: the `==` results were **never asserted**, so the test validated nothing.

## Fix

- Replace `pytest.approx` on datetimes with plain `abs(a - b) <= timedelta(seconds=30)` assertions — correct on every pytest version, and now actually asserted.
- Sync the RTC to the runner clock at test start so the "previous time within 30 s of now" assertion holds on a fresh boot (the RTC powers up at 2000-01-01 — visible as `1999-12-31` EVR timestamps in CI logs before this test runs).
- Remove the copy-paste block that re-fetched the same `TimeSet` event and compared a stale variable against "now".

## Related, not addressed here

- The `await_event` calls in tests 05–10 never assert (they return `None` on timeout), and test_10 leaves an alarm armed — PR #407 already covers that ground.
- Pinning CI Python deps (PR #431) would prevent this class of drift entirely.

### DIFF
--- a/PROVESFlightControllerReference/test/int/rtc_test.py
+++ b/PROVESFlightControllerReference/test/int/rtc_test.py
@@ -102,6 +102,11 @@ def uplink_sequence_and_await_completion(
 def test_01_time_set(fprime_test_api: IntegrationTestAPI, start_gds):
     """Test that we can set the time"""
 
+    # Sync the RTC to the test runner's clock first so the "previous time"
+    # reported by the next TIME_SET is known (a fresh boot starts at 2000-01-01)
+    set_time(fprime_test_api)
+    fprime_test_api.clear_histories()
+
     # Set time to Curiosity landing on Mars (7 minutes of terror! https://youtu.be/Ki_Af_o9Q9s)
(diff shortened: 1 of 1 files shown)


nateinaction:
Already working on this in #433

PR #440, opened 2026-07-12
Add TelemetryArchive Component

### LINKED ISSUE #423: [FEATURE] Store-and-forward telemetry
### Short summary

Buffer telemetry (and possibly payload data) to filesystem when it cannot be downlinked, then replay it when a link is available. Look at risk to the filesystem.

### Motivation and user story

As an operator, I want telemetry recorded during comms outages, so that sensor and payload data is not lost between passes.

### Proposed implementation

Today telemetry flows through CdhCore.tlmSend → comSplitterTelemetry → LoRa/UART queues. If the radio is off, the queue is full, or downlink is rate-limited (telemetryDelay, downlinkDelay), packets are dropped. 

FsSpace reports free space on a FAT32 partition

Design decisions 

What to store

- undownlinked spacecraft telemetry packets would be useful, especially the first few on a boot (especially first boot)
- SatNOGS/payload telemetry would be useful for BroncoSpace

Other Questions
Write rate — per-packet, batched, or on a fixed interval?
Read/forward CONOPS — how do we get the command to downlink, how do we downlink one piece of telem at a time?
Storage budget — max file size, max total usage
Filesystem safety — bounded writes, wear awareness, behavior when FsSpace.FreeSpace is low. We cannot risk the filesystem filling or corrupting 

### Priority

Medium

hrfarmer:
## Summary
- Add a new `TelemetryArchive` component that captures packetized telemetry into bounded deployment and rolling filesystem segments.
- Persist archive state with segment rotation, capacity checks, pruning for rolling data, and status telemetry/events.
- Extend `AntennaDeployer` to report persistence success and emit a completion signal when deployment state is safely stored.
- Wire the new archive into the reference topology and telemetry packet definitions.


hrfarmer:
rewriting in a new branch

PR #441, opened 2026-07-13
Add repository and submodule reset step

hrfarmer:
Adds a step in the build workflow to reset the repository and submodules to their original states to ensure changes from other prs don't linger


Mikefly123 (approved):
LGTM!

PR #455, opened 2026-07-20
Timebase paramter

collinc04:
# set_timebase

## Description

Added a command to switch between timebases

## Interactions

Does not interfere with command loss timer as that has been moved to the 10hz rate group.
Cancels all queued sequences, just like time_set does

Does not interact with any of the other RTC manager commands, only changes the behavior of the time get port

## How Has This Been Tested?

- [ ] Unit tests
- [x] Integration tests
- [ ] Z Tests
- [ ] Manual testing (describe steps)

## Checklist


nateinaction (changes requested):
I like this initial take, here's a suggested design:

Use F Prime's built-in enum `TimeBase`. Store the selected timebase as an F Prime parameter so it can be stored over multiple boots. No F Prime commands necessary, the parameter will be all you need to set and store the chosen timebase. The parameter default should be `TB_SC_TIME`. Check the timebase once near the beginning of `timeGetPort_handler()` after the variable sets but before the RTC device ready check. If the timebase is `TB_PROC_TIME`, return the proc time.

Also to consider... Since this component supports fewer time bases than the upstream F Prime `TimeBase` enum we may want to have our own `TimeBase` enum declared in the FPP that only copies the upstream enum but removes the unsupported timebases like `TB_WORKSTATION_TIME`.

What do you think?

collinc04:
> I like this initial take, here's a suggested design:
> 
> Use F Prime's built-in enum `TimeBase`. Store the selected timebase as an F Prime parameter so it can be stored over multiple boots. No F Prime commands necessary, the parameter will be all you need to set and store the chosen timebase. The parameter default should be `TB_SC_TIME`. Check the timebase once near the beginning of `timeGetPort_handler()` after the variable sets but before the RTC device ready check. If the timebase is `TB_PROC_TIME`, return the proc time.
> 
> Also to consider... Since this component supports fewer time bases than the upstream F Prime `TimeBase` enum we may want to have our own `TimeBase` enum declared in the FPP that only copies the upstream enum but removes the unsupported timebases like `TB_WORKSTATION_TIME`.
> 
> What do you think?

I like the idea of using a timebase to survive multiple boots, I think this is a good idea

PROVESFlightControllerReference/ReferenceDeployment/Top/ReferenceDeploymentPackets.fppi
```diff
@@ -11,6 +11,7 @@ telemetry packets ReferenceDeploymentPackets {
 
   packet Beacon id 1 group 1 {
     # Satellite Meta Data
+    ReferenceDeployment.lora.ModuleMode
```
Mikefly123: I'm pretty sure this is not an actual telemetry channel that is available in the LoRa component. It would be nice if the upstream component had something like this though!

Mikefly123 (dismissed):

PROVESFlightControllerReference/ReferenceDeployment/Top/ReferenceDeploymentPackets.fppi
collinc04: I added one as a pr on the submodule, it needs to be changed in order to be more descriptive though. It hasn't been accepted yet and that's why build fails on this PR at the moment

collinc04:
I'd like to note that I have separated this from the beaconing packet change, work will continue on that one in the fprime-zephyr repo

lib/fprime-zephyr
ineskhou, line 1: Can you update the submodules to match the ones on main?

ineskhou, line 1: There's a piece on the ticket:

Audit time usage Document where time matters and what TIME_SET changes, including:

Command sequences (cmdSeq, payloadSeq, safeModeSeq) — canceled on time set
AuthenticationRouter comm-loss timer (switched by another PR)

I think Detumble also falls into that category. Can you audit and explain what happens to all the components that depend on time when this time set hits?

PROVESFlightControllerReference/Components/Drv/RtcManager/RtcManager.cpp
```diff
@@ -48,9 +49,17 @@ void RtcManager ::timeGetPort_handler(FwIndexType portNum, Fw::Time& time) {
...
+    if (this->m_ProcTimeSet) {
+        // use proc time
+        time.set(TimeBase::TB_PROC_TIME, 0, seconds_since_boot, useconds_since_boot);
+        return;
+    }
+
     // Check device readiness
     if (!device_is_ready(this->m_dev)) {
         this->log_CONSOLE_RtcNotReady();
+        this->m_ProcTimeSet = true;  // proc time flag
```
ineskhou: My vision for this component was more something that didn't change the behavior of the RTC bring up, but something the operator could physcially comamnd. That way we don't have to audit the behavoir of every failing case and how this flag would change it/ less changes are always better. 

That being said totally open to it, if tested for these cases. It could fix issues if we've been having timebase issues with flashing back and forth between timebases, and I think if our sat completly runs on monotonic time, that would be fine and safer than a flaky RTC.

I think the other thing is, the RTC does fail on the first few seconds of startup, you can observe monotonic time for a bit before the RTC kicks on. Do these cases that do not allow the RTC to retry account for that or does every run run on monotinic time?

PROVESFlightControllerReference/Components/Drv/RtcManager/RtcManager.fpp
```diff
@@ -31,6 +31,9 @@ module Drv {
             t: Drv.TimeData @< Set the time
         )
 
+        @ TO_PROC_TIME command to use proc time instead of RTC time
+        sync command TO_PROC_TIME()
```
ineskhou: Could be good to have a testcase for this commadn to show we can toggle it on and off no problem. We could push back getting the test in after the code freeze, but we should make a ticket if thats the case

ineskhou (changes requested):

lib/fprime-zephyr
collinc04: Yes I'll do that right now, I thought they were the same. I must have pulled the wrong commit

PROVESFlightControllerReference/Components/Drv/RtcManager/RtcManager.fpp
collinc04: I made one, forgot to push oops :sweat:

collinc04: Could use some more verification though? It only checks for event emission, what do you think?

collinc04: This also uses await, so It probably needs to be changed to be more in line with #407

PROVESFlightControllerReference/test/int/rtc_test.py
```diff
@@ -436,3 +436,20 @@ def test_10_double_set_test(fprime_test_api: IntegrationTestAPI, start_gds):
...
+    curiosity_landing = datetime(2012, 8, 6, 5, 17, 57, tzinfo=timezone.utc)
+    set_time(fprime_test_api, curiosity_landing)
+
+    # Assert that we receive a TimeBase event within 10 seconds
+    fprime_test_api.await_event(f"{rtcManager}.TimeBase", timeout=10)
+
+    fprime_test_api.send_command(f"{rtcManager}.TO_PROC_TIME")
+
+    # Assert that we receive a TimeBase event within 10 seconds
+    fprime_test_api.await_event(f"{rtcManager}.TimeBase", timeout=10)
```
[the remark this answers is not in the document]
collinc04: The event emissions already trigger at the same time, so we can assume we got a response from TimeBase as well. I'm not entirely sure this is necessary

PROVESFlightControllerReference/Components/Drv/RtcManager/RtcManager.cpp
collinc04: My general thought was that if we implement a command toggle for it, then it should stay that way if it fails unless commanded otherwise so that we don't switch back to a potentially bad RTC. 

I did remove the flag set on the device not ready case in the bring up, because otherwise it would boot up in monotonic time unless otherwise set

PROVESFlightControllerReference/test/int/rtc_test.py
ineskhou: Could be good to have this test assert what the timebase is that is being set (PROC vs RTC) and/or actually check the physcial timebase. Checking the event doesn't let us know if the system of the sat is on PROC time. At minimum before merge, woudl want manual evidence on this happening on a board 

That being said, test updates will not block merge.

PROVESFlightControllerReference/test/int/rtc_test.py
```diff
@@ -436,3 +436,20 @@ def test_10_double_set_test(fprime_test_api: IntegrationTestAPI, start_gds):
     fprime_test_api.send_command(f"{rtcManager}.ALARM_SET", [alarm_time_data_str])
     # Assert that we receive an AlarmNotSet event within 10 seconds
     fprime_test_api.await_event(f"{rtcManager}.AlarmNotSet", timeout=10)
+
+
+@pytest.mark.uart_only(reason="Test functionality of to_proc_time")
+def test_proc_toggle(fprime_test_api: IntegrationTestAPI, start_gds):
+    """Test for events emitted by proc time toggle"""
+
+    # Set time to Curiosity landing on Mars (7 minutes of terror! https://youtu.be/Ki_Af_o9Q9s)
```
ineskhou: love it!!!

PROVESFlightControllerReference/Components/Drv/RtcManager/RtcManager.fpp
```diff
@@ -55,6 +58,16 @@ module Drv {
             useconds: U32 @< Microseconds
         ) severity activity high id 3 format "Time set on RTC, previous time: {}.{}"
 
+        @ ProcTimeSet event indicates that proc time was set
+        event ProcTimeSet(
+            seconds: U32 @< Uptime in seconds
+        ) severity activity high id 17 format "Proc time set, current uptime: {}"
+
+        @ Timebase even fires when the timebase changes and indicates the current timebase
```
ineskhou: Typo? Even --> event

PROVESFlightControllerReference/Components/Drv/RtcManager/RtcManager.cpp
ineskhou: But what if we have flaky events? one wrong I2C value could set us to monotonic. I also think the behavoir of the RTC is that we start on monotonic on boot, so if the RTC is not working we are on monotonic time. At least that's what I've observe on my board in the past.

If you want to keep latched implementation of the component, I think we at least want events in these places so that we know why the time is set to monotonic. I'd also want that more visible on the sdd

lib/fprime-zephyr
ineskhou: Following up on this one, I think we need to cancel running sequences when we set the time, otherwise unexpected things may happen.

I also have a board without an RTC battery so am unable to test these toggle changes. Do you mind uploading verification evidence like I did on the hardcoded startup component?

PROVESFlightControllerReference/Components/Drv/RtcManager/RtcManager.cpp
```diff
@@ -151,13 +157,30 @@ void RtcManager ::TIME_SET_cmdHandler(FwOpcodeType opCode, U32 cmdSeq, Drv::Time
...
+    // Switch the time source to proc time
+    Fw::Time proc_time;
+    int64_t t = k_uptime_get();
+    U32 seconds_since_boot = static_cast<U32>(t / 1000);
+    U32 useconds_since_boot = static_cast<U32>((t % 1000) * 1000);
+    this->log_ACTIVITY_HI_procTimeSet(seconds_since_boot);  // Log the current uptime in seconds
+
+    this->m_ProcTimeSet = true;  // proc time flag
+    proc_time.set(TimeBase::TB_PROC_TIME, 0, seconds_since_boot, useconds_since_boot);
+
```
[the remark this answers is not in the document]
ineskhou: I think I agree with this one, I'm not sure what the proc_time.set does here

ineskhou (dismissed):
Sorry hope these comments aren't too much of a mess bascially:

1. Up to you as designer to choose to latch to monotonic for errors or not (you can also ask for a trade in the discord if you want), but if you do I'd want more clear events and documentation for it in the sdd.
2. Looking for manual testing (see startup manager)
3. Want checking that other parts of the system that use the RTC will not be negatavly affected by these timebase  changes.
4. Few nits (spelling, extra variable, ect)

But you've put a lot of work in this and it looks great! Thanks for staying in there through all these reviews

PROVESFlightControllerReference/Components/Drv/RtcManager/RtcManager.cpp
```diff
@@ -78,13 +88,16 @@ void RtcManager ::timeGetPort_handler(FwIndexType portNum, Fw::Time& time) {
...
         this->log_CONSOLE_RtcInvalidTime();
+        this->m_ProcTimeSet = true;  // proc time flag
 
         // Use uptime as fallback
         time.set(TimeBase::TB_PROC_TIME, 0, seconds_since_boot, useconds_since_boot);
         return;
     }
     this->log_CONSOLE_RtcInvalidTime_ThrottleClear();
 
+    this->m_ProcTimeSet = false;  // proc time flag
```
ineskhou: Not sure if we need this extra set

ineskhou (commented):
Looks good! Just waiting on some testing n docs, few smaller comments

PROVESFlightControllerReference/Components/Drv/RtcManager/RtcManager.cpp
```diff
@@ -151,13 +160,37 @@ void RtcManager ::TIME_SET_cmdHandler(FwOpcodeType opCode, U32 cmdSeq, Drv::Time
         return;
     }
 
+    this->m_ProcTimeSet = false;  // proc time flag
+
```
nateinaction: I think I understand the motive here and it's a good one. If the operator sets the time then surely they want to use the RTC? To me, this feels like a side effect. Let's not change the time base because we set the time. Let's have the operator control these separately and chain them together if that's the behavior they want.

PROVESFlightControllerReference/Components/Drv/RtcManager/RtcManager.cpp
```diff
@@ -151,13 +160,37 @@ void RtcManager ::TIME_SET_cmdHandler(FwOpcodeType opCode, U32 cmdSeq, Drv::Time
...
     // Emit time set event, include previous time for reference
     this->log_ACTIVITY_HI_TimeSet(time_before_set.getSeconds(), time_before_set.getUSeconds());
+    this->log_ACTIVITY_HI_TimeBase(Fw::String("RTC"));
```
nateinaction: Same note.

PROVESFlightControllerReference/Components/Drv/RtcManager/RtcManager.fpp
```diff
@@ -55,6 +58,16 @@ module Drv {
...
         ) severity activity high id 3 format "Time set on RTC, previous time: {}.{}"
 
+        @ ProcTimeSet event indicates that proc time was set
+        event ProcTimeSet(
+            seconds: U32 @< Uptime in seconds
+        ) severity activity high id 17 format "Proc time set, current uptime: {}"
+
+        @ Timebase event fires when the timebase changes and indicates the current timebase
+        event TimeBase(
+            timeBase: string size 4
```
nateinaction: See if you can use the time base type from upstream F Prime.

PROVESFlightControllerReference/Components/Drv/RtcManager/RtcManager.cpp
```diff
@@ -151,13 +160,37 @@ void RtcManager ::TIME_SET_cmdHandler(FwOpcodeType opCode, U32 cmdSeq, Drv::Time
...
+        if (!this->isConnected_cancelSequences_OutputPort(i)) {
+            continue;
+        }
+        this->cancelSequences_out(i);
+    }
+
+    // Switch the time source to proc time
+    int64_t uptime = k_uptime_get();
+    U32 seconds_since_boot = static_cast<U32>(uptime / 1000);
+    this->log_ACTIVITY_HI_ProcTimeSet(seconds_since_boot);  // Log the current uptime in seconds
```
nateinaction: I don't think it is necessary to send an event with proctime because as soon as you send the activity on 188, the new time base will appear with the event.

PROVESFlightControllerReference/Components/Drv/RtcManager/RtcManager.fpp
nateinaction: Let's take a time base argument in the command.

nateinaction (changes requested):
This is much simpler than before. Great job!

I'd like to challenge you to use F Prime's types instead of strings and bools. Using F Prime's type system will help you build resilient and reusable components. My previous review suggests a way to build out this feature using F Prime's types:

> Use F Prime's built-in enum TimeBase. Store the selected timebase as an F Prime parameter so it can be stored over multiple boots. No F Prime commands necessary, the parameter will be all you need to set and store the chosen timebase. The parameter default should be TB_SC_TIME. Check the timebase once near the beginning of timeGetPort_handler() after the variable sets but before the RTC device ready check. If the timebase is TB_PROC_TIME, return the proc time.
> 
> Also to consider... Since this component supports fewer time bases than the upstream F Prime TimeBase enum we may want to have our own TimeBase enum declared in the FPP that only copies the upstream enum but removes the unsupported timebases like TB_WORKSTATION_TIME.

PROVESFlightControllerReference/Components/Drv/RtcManager/RtcManager.cpp
collinc04: Roger :saluting_face: My initial thought was that doing it like this would make switching timebases more intuitive but I can see how that restricts operators

collinc04:
Changing this to work with enumeration and save accross boot, currently broken on local but other than surviving a reboot, behavior won't change from this commit. Let me know if that seems right

PROVESFlightControllerReference/Components/Drv/RtcManager/RtcManager.cpp
ineskhou: But wouldn't you want a way to debug that the event actually went through to the sat? I think more visibility is better in this case

PROVESFlightControllerReference/Components/Drv/RtcManager/RtcManager.cpp
```diff
@@ -158,6 +167,65 @@ void RtcManager ::TIME_SET_cmdHandler(FwOpcodeType opCode, U32 cmdSeq, Drv::Time
...
+        this->cancelSequences_out(i);
+    }
+
+    // Retrieve time and it's associated base enum
+    FwTimeBaseStoreType curr_time_base = this->getTime().getTimeBase();
+
+    bool valid = false;
+
+    // Check if TimeBase is already correct
+    if (tb == curr_time_base) {
```
ineskhou: Say we sent a proctime command but its at the begining of startup, so the time has not gone from proctime to RTC time like generall happens. What would happen when the RTC time kicks in?

I think you can't make the assumption that just because the current timebase is a timebase that the flag is also set to that

PROVESFlightControllerReference/Components/Drv/RtcManager/RtcManager.cpp
```diff
@@ -158,6 +167,65 @@ void RtcManager ::TIME_SET_cmdHandler(FwOpcodeType opCode, U32 cmdSeq, Drv::Time
...
 }
 
+void RtcManager ::SET_TIMEBASE_cmdHandler(FwOpcodeType opCode, U32 cmdSeq, FwTimeBaseStoreType tb) {
+    // Cancel any running sequences before setting time, as time change may impact their behavior
+    for (FwIndexType i = 0; i < this->getNum_cancelSequences_OutputPorts(); i++) {
+        if (!this->isConnected_cancelSequences_OutputPort(i)) {
+            continue;
+        }
+        this->cancelSequences_out(i);
+    }
```
ineskhou, line 186: You cancel sequences before validating the timebase change which means that an invalid timebase would cancel sequences for no reason. I think that is okay but the logic makes more sense the other way around

PROVESFlightControllerReference/test/int/rtc_test.py
```diff
@@ -436,3 +436,23 @@ def test_10_double_set_test(fprime_test_api: IntegrationTestAPI, start_gds):
...
+
+    # Test that we can set timebase to RTC time
+    fprime_test_api.send_command(f"{rtcManager}.SET_TIMEBASE", [3])
+    # Assert that we receive a TimeBase event within 10 seconds
+    fprime_test_api.await_event(f"{rtcManager}.TimeBase", timeout=10)
+
+    # Test that invalid times are filtered out
+    fprime_test_api.send_command(f"{rtcManager}.SET_TIMEBASE", [0])
+    # Assert that we receive a TimeBase event within 10 seconds
+    fprime_test_api.await_event(f"{rtcManager}.TimeBase", timeout=10)
```
ineskhou: Do we get a get a timebase event on an invalid event?

PROVESFlightControllerReference/Components/Drv/RtcManager/RtcManager.cpp
```diff
@@ -158,6 +167,65 @@ void RtcManager ::TIME_SET_cmdHandler(FwOpcodeType opCode, U32 cmdSeq, Drv::Time
...
+
+        this->log_ACTIVITY_HI_TimeBase(tb);
+        valid = true;
+
+        // Command response
+        this->cmdResponse_out(opCode, cmdSeq, Fw::CmdResponse::OK);
+        return;
+    }
+
+    if (!valid) {
```
ineskhou: Do you need the valid flag, if all the valid paths return early anyways

PROVESFlightControllerReference/Components/Drv/RtcManager/RtcManager.cpp
```diff
@@ -158,6 +168,18 @@ void RtcManager ::TIME_SET_cmdHandler(FwOpcodeType opCode, U32 cmdSeq, Drv::Time
...
+    // Cancel any running sequences before setting time, as time change may impact their behavior
+    for (FwIndexType i = 0; i < this->getNum_cancelSequences_OutputPorts(); i++) {
+        if (!this->isConnected_cancelSequences_OutputPort(i)) {
+            continue;
+        }
+        this->cancelSequences_out(i);
+    }
+
+    this->log_ACTIVITY_HI_TimeBaseSwitch(id);
+}
```
nateinaction, line 189: Two issues here:

1. **The event payload is wrong.** `TimeBaseSwitch` declares `timeBase: FwTimeBaseStoreType`, but this passes `id`, the parameter ID. `PARAMID_TIMEBASE` is `0`, so the downlinked event always reads `Timebase switched to: 0` whichever direction you switch. The integration test doesn't catch it because it only checks that the event fired, not its payload.
2. **No parameter-ID guard.** `parameterUpdated` fires for every parameter on the component. With one parameter that's harmless, but the moment a second is added this cancels sequences and emits a timebase event on an unrelated update.

Suggested:

```suggestion
void RtcManager ::parameterUpdated(FwPrmIdType id) {
    if (id != RtcManager::PARAMID_TIMEBASE) {
        return;
    }

    Fw::ParamValid valid;
    const Rtc::TimeBase timeBase = this->paramGet_TIMEBASE(valid);
    if ((valid == Fw::ParamValid::INVALID) || (valid == Fw::ParamValid::UNINIT)) {
        return;
    }

    // Cancel any running sequences, as the change in reported time may impact their behavior
    for (FwIndexType i = 0; i < this->getNum_cancelSequences_OutputPorts(); i++) {
        if (!this->isConnected_cancelSequences_OutputPort(i)) {
            continue;
        }
        this->cancelSequences_out(i);
    }

    this->log_ACTIVITY_HI_TimeBaseChanged(timeBase);
}
```

(The `TimeBaseChanged` name assumes the rename suggested on the `.fpp`; keep `TimeBaseSwitch` if you'd rather not rename.)

PROVESFlightControllerReference/Components/Drv/RtcManager/RtcManager.fpp
```diff
@@ -55,6 +68,11 @@ module Drv {
             useconds: U32 @< Microseconds
         ) severity activity high id 3 format "Time set on RTC, previous time: {}.{}"
 
+        @ Timebase event fires when the timebase changes and indicates the current timebase
+        event TimeBaseSwitch(
+            timeBase: FwTimeBaseStoreType
+        ) severity activity high id 18 format "Timebase switched to: {}"
```
nateinaction: Typing the payload as the raw `FwTimeBaseStoreType` downlinks a bare integer — the GDS will show `Timebase switched to: 1` rather than the enum name. Since you already have `Rtc.TimeBase`, use it and get a rendered symbol for free.

I'd also rename `TimeBaseSwitch` → `TimeBaseChanged`: the event reports the resulting state, not an action, and it lines up with the `parameterUpdated` hook that emits it. Renaming touches `RtcManager.cpp` and `rtc_test.py` too — drop the rename if you'd rather keep the diff tight, but the type change is worth taking either way.

```suggestion
        @ TimeBaseChanged event fires when the timebase parameter changes and indicates the new timebase
        event TimeBaseChanged(
            timeBase: Rtc.TimeBase @< The timebase now in use
        ) severity activity high id 18 format "Timebase changed to: {}"
```

PROVESFlightControllerReference/Components/Drv/RtcManager/RtcManager.cpp
```diff
@@ -48,9 +48,19 @@ void RtcManager ::timeGetPort_handler(FwIndexType portNum, Fw::Time& time) {
...
     U32 useconds_since_boot = static_cast<U32>((t % 1000) * 1000);
 
+    // check if proc time mode is set
+    Fw::ParamValid is_valid;
+    FwTimeBaseStoreType curr_tb = this->paramGet_TIMEBASE(is_valid);
+    if (curr_tb == Rtc::TimeBase::TB_PROC_TIME) {
+        // use proc time
+        time.set(TimeBase::TB_PROC_TIME, 0, seconds_since_boot, useconds_since_boot);
+        return;
+    }
```
nateinaction: `paramGet_TIMEBASE` returns `Rtc::TimeBase`, not `FwTimeBaseStoreType` — declaring `curr_tb` as the storage type relies on the implicit conversion and loses the type that makes the comparison below meaningful. I still don't think the prefix `::` fully matters but I looked into it further and I believe it prevents ambiguity between compilers which may not know whether to use the global `TimeBase` or the `Timebase` in the `Rtc` class so let's use `::TimeBase` now that `Rtc::TimeBase` is in scope.

```suggestion
    // Use proc time directly when the timebase parameter selects it
    Fw::ParamValid timeBaseValid;
    const Rtc::TimeBase timeBase = this->paramGet_TIMEBASE(timeBaseValid);
    if (timeBase == Rtc::TimeBase::TB_PROC_TIME) {
        time.set(::TimeBase::TB_PROC_TIME, 0, seconds_since_boot, useconds_since_boot);
        return;
    }
```

PROVESFlightControllerReference/Components/Drv/RtcManager/RtcManager.cpp
```diff
@@ -48,9 +48,19 @@ void RtcManager ::timeGetPort_handler(FwIndexType portNum, Fw::Time& time) {
...
+
     // Check device readiness
     if (!device_is_ready(this->m_dev)) {
         this->log_CONSOLE_RtcNotReady();
+        // Don't set the proc time flag if device is not ready
```
nateinaction: This comment contradicts the line right below it, which does set `TB_PROC_TIME` — that's the intended fallback when the RTC isn't ready, and it predates this PR. Suggest dropping the comment.

```suggestion
```

PROVESFlightControllerReference/test/int/rtc_test.py
```diff
@@ -436,3 +436,22 @@ def test_10_double_set_test(fprime_test_api: IntegrationTestAPI, start_gds):
...
+    fprime_test_api.assert_event(
+        f"{rtcManager}.TimeBaseSwitch", start="NOW", timeout=10
+    )
+
+    # Test that we can set timebase to RTC time
+    fprime_test_api.send_command(f"{rtcManager}.TIMEBASE_PRM_SET", ["TB_SC_TIME"])
+    # Assert that we receive a TimeBase event within 10 seconds
+    fprime_test_api.assert_event(
+        f"{rtcManager}.TimeBaseSwitch", start="NOW", timeout=10
+    )
```
nateinaction: Three problems with this test as written:

1. **Race.** `send_command` returns as soon as the command is dispatched, so `assert_event(start="NOW")` can open its window *after* the event has already arrived and then time out. `await_event` searches the existing history and is the pattern used by the rest of this file. Take a look at other integration tests which store a `start_time` var prior to running commands to get an idea of how to better handle this.
2. **Command rejection goes unnoticed.** Plain `send_command` doesn't assert the command response; if `TIMEBASE_PRM_SET` is rejected the test only fails later, on a confusing event timeout. `proves_send_and_assert_command` (already imported at the top of this file) covers both.
3. **No cleanup on failure.** If the first assertion throws, the spacecraft is left on `TB_PROC_TIME` and every subsequent test in the session sees uptime-based timestamps instead of RTC time. Wrapping the restore in `finally` keeps the failure contained to this test.

Here is a partial suggestion, excluding a fix for 1.

```suggestion
@pytest.mark.uart_only(reason="Test functionality of the timebase parameter")
def test_11_proc_toggle(fprime_test_api: IntegrationTestAPI, start_gds):
    """Test for events emitted by the timebase parameter"""

    try:
        # Test that we can set timebase to proc time
        proves_send_and_assert_command(
            fprime_test_api, f"{rtcManager}.TIMEBASE_PRM_SET", ["TB_PROC_TIME"]
        )
        # Assert that we receive a TimeBaseChanged event within 10 seconds
        fprime_test_api.await_event(f"{rtcManager}.TimeBaseChanged", timeout=10)
    finally:
        # Restore spacecraft time so subsequent tests see RTC-backed timestamps
        proves_send_and_assert_command(
            fprime_test_api, f"{rtcManager}.TIMEBASE_PRM_SET", ["TB_SC_TIME"]
        )
        fprime_test_api.await_event(f"{rtcManager}.TimeBaseChanged", timeout=10)
```

PROVESFlightControllerReference/Components/Drv/RtcManager/docs/sdd.md
```diff
@@ -115,6 +117,11 @@ This logic applies both when using the RTC (`TB_SC_TIME`) and when in failover m
 | ALARM_CANCEL | Cancels the current alarm |
 | ALARM_LIST | Responds with info about the current set alarm |
 
+## Parameters
+| Name | Description |
+|---|---|
+| TIMEBASE | Decides the timebase that timeGetPort reports |
```
nateinaction: Worth expanding the docs a bit while you're here — a reader can't tell from this table what the parameter's type, default, or persistence behavior is, or what happens on `timeGetPort` when it's set to `TB_PROC_TIME`. Things I'd add:

- Type and default in this table (`Rtc.TimeBase`, `TB_SC_TIME`), plus a note that `TIMEBASE_PRM_SAVE` is what persists it across boots.
- A `TimeBaseChanged` row in the Events table below.
- A line in the `timeGetPort` usage section noting it returns uptime without touching the RTC when the parameter is `TB_PROC_TIME`.
- A note on why the enum lives in its own `Rtc` module rather than `Drv` (a `Drv.TimeBase` would shadow the global `TimeBase` inside generated `Drv` code) — that's non-obvious enough that someone will otherwise try to "fix" it.
- The `parameterUpdated` entry in the class diagram, and a short sequence diagram for the parameter set/save flow to match the other commands.

nateinaction (commented):
Nice work! Moving this from a command to a persisted parameter. I caught a few remaining issues but this is definitely the right shape!

- The PR description still says "Added a command to switch between timebases"; worth updating to describe the parameter (and the `TIMEBASE_PRM_SAVE` step needed to persist across boots).

nateinaction (approved):

PR #464, opened 2026-07-21
TlmArchive

hrfarmer:
This component will store the telemetry collected before the first antenna deploy, as this information can be very useful for early troubleshooting.

This component was scoped down a lot from what I originally intended, it was functional but writes/other fs actions kept occasionally causing a software hang, which is very undesirable. I'll keep working on an expansion to this component that can constantly log telemetry so any recent missing telemetry could be downlinked, and if the software hangs can be fixed, could go on the second partition.

- [x] Update SDD


lib/fprime-zephyr
ineskhou, line 1: I think you need to pull the latest zephyr submoduels

PROVESFlightControllerReference/Components/TlmArchive/docs/sdd.md
```diff
@@ -0,0 +1,5 @@
+# TlmArchive
+
+`TlmArchive` appends telemetry packets to `//tlm/pre_deployment.tlm`
+while the antenna deployment state is false. Once the antenna is marked
+deployed, incoming telemetry is no longer written.
```
ineskhou: If you could list the ports and events related to this component here to match other sdds that would be great!

PROVESFlightControllerReference/Components/AntennaDeployer/AntennaDeployer.fpp
```diff
@@ -79,6 +82,9 @@ module Components {
         @ Port receiving calls from the rate group
         sync input port schedIn: Svc.Sched
 
+        @ Port returning whether antenna deployment has completed
+        sync input port deploymentStateGet: Components.GetDeploymentState
+
         @ Port signaling the burnwire component to start heating
         output port burnStart: Fw.Signal
```
ineskhou, line 89: Could we add events here so we know when the writes stop and start?

It would also help a lot with testing and assurance

PROVESFlightControllerReference/Components/TlmArchive/TlmArchive.cpp
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    const FwSizeType requestedSize = packet.getSize();
+    FwSizeType writtenSize = requestedSize;
+    const Os::File::Status writeStatus = file.write(packet.getBuffAddr(), writtenSize);
+    if ((writeStatus != Os::File::OP_OK) || (writtenSize != requestedSize)) {
+        this->log_WARNING_HI_ArchiveWriteError(Os::FileStatus(static_cast<Os::FileStatus::T>(writeStatus)),
+                                               requestedSize, writtenSize);
+    }
+    file.close();
```
ineskhou, line 221: Its possible that antenna deploy does not run for whatever reason (startup sequence is not there, seauencer doesnt run, rad hit...) We dont have any way that stops writing other than that event. Can we only write like a max number to the file or something like that?

PROVESFlightControllerReference/Components/TlmArchive/TlmArchive.cpp
```diff
@@ -0,0 +1,81 @@
...
+        this->log_WARNING_HI_ArchiveFileError(Fw::LogStringArg("open_append"));
+        return;
+    }
+
+    const FwSizeType requestedSize = packet.getSize();
+    FwSizeType writtenSize = requestedSize;
+    const Os::File::Status writeStatus = file.write(packet.getBuffAddr(), writtenSize);
+    if ((writeStatus != Os::File::OP_OK) || (writtenSize != requestedSize)) {
+        this->log_WARNING_HI_ArchiveWriteError(Os::FileStatus(static_cast<Os::FileStatus::T>(writeStatus)),
+                                               requestedSize, writtenSize);
```
ineskhou: If the filesystem is unhealthy I think we would get spammed with these warnings, which would be bad for ever getting telemtry beacons. Can we throttle these/file system attempts? I think hammering the file system is also not ideal.

PROVESFlightControllerReference/Components/TlmArchive/TlmArchive.cpp
```diff
@@ -0,0 +1,81 @@
...
+    if (!this->m_antennasDeployed) {
+        this->m_antennasDeployed = this->deploymentStateGet_out(0);
+    }
+
+    if (this->m_antennasDeployed) {
+        return;
+    }
+
+    if (Os::FileSystem::createDirectory(TLM_DIRECTORY, false) != Os::FileSystem::OP_OK) {
+        this->log_WARNING_HI_ArchiveFileError(Fw::LogStringArg("create_directory"));
```
ineskhou: I also worry trying to create a directory on every cycle on the 1 Hz rate group. Its a lot of file operations to do in 1Hz espcieally repeatably

ineskhou (changes requested):
Great stuff! Great use of topology and ports and great architechure design. Had some comments about robustness with the file systsem since we dont want to hammer the file system too hard when we are depending on teh startup sequence

Main thing I want to see before merge is some manual testing of this, if you could provide screenshots. Especially this working as the startup sequence runs and if you can recerate a dead filesystem (but no worries if not)

ineskhou (changes requested):
Great stuff! Great use of topology and ports and great architechure design. Had some comments about robustness with the file systsem since we dont want to hammer the file system too hard when we are depending on teh startup sequence

Main thing I want to see before merge is some manual testing of this, if you could provide screenshots. Especially this working as the startup sequence runs and if you can recerate a dead filesystem (but no worries if not)

PROVESFlightControllerReference/Components/TlmArchive/TlmArchive.cpp
hrfarmer: agree, will change it to only create once

PROVESFlightControllerReference/Components/TlmArchive/TlmArchive.cpp
hrfarmer: I think instead of throttling the event I'll just have it stop trying to write if it fails to write a few times, since that would probably mean something is wrong with the fs.

lib/fprime-zephyr
hrfarmer: done

PROVESFlightControllerReference/Components/TlmArchive/TlmArchive.cpp
hrfarmer: added a max file size limit (currently 10000 bytes)

PROVESFlightControllerReference/Components/TlmArchive/docs/sdd.md
hrfarmer: will update the sdd as a final step once everything else in the component is good 👍

PROVESFlightControllerReference/Components/AntennaDeployer/AntennaDeployer.fpp
hrfarmer: added them (does add a bit of event spam, but might not matter since those would be outputting only before transmit is enabled)

PROVESFlightControllerReference/Components/TlmArchive/TlmArchive.cpp
hrfarmer: added a filesystem failure limit before it stops attempting writes (currently at 3)

PROVESFlightControllerReference/Components/TlmArchive/TlmArchive.fpp
```diff
@@ -0,0 +1,45 @@
...
+            requested: FwSizeType @< Requested byte count
+            written: FwSizeType @< Reported byte count
+        ) severity warning high \
+          format "Pre-deployment telemetry archive write failed: status {}, requested {}, wrote {}"
+
+        @ Reports when telemetry archiving is disabled due to hitting the failure limit
+        event ArchiveWriteDisabled(
+            count: I8
+            maxSize: FwSizeType
+        ) severity warning high format "{} filesystem failures counted and/or {}b file size limit reached; disabling further telemetry writes." throttle 1
```
ineskhou: I feel like filesystem limit and filesystem fails shold be 2 different events, since it might be confusing for an operator if we only get one event and no context for the amount of errors and filesteme limt

ineskhou: I also feel like adding an overfoling file to m_failures maybe deosnt make as much sense

PROVESFlightControllerReference/Components/TlmArchive/TlmArchive.cpp
```diff
@@ -0,0 +1,137 @@
...
+            this->m_failures++;
+        }
+        return;
+    }
+    this->m_directoryInitialized = true;
+
+    const FwSizeType requestedSize = packet.getSize();
+    FwSizeType currentSize = 0;
+    const Os::FileSystem::Status sizeStatus = Os::FileSystem::getFileSize(PRE_DEPLOYMENT_TLM_PATH, currentSize);
+    if ((sizeStatus != Os::FileSystem::OP_OK) && (sizeStatus != Os::FileSystem::DOESNT_EXIST)) {
```
ineskhou: In the interest of having less file operations on every since 1Hz rate group could we use m_file_size instead of calling the Os function every time? Less file actions per rate group the more robsut

ineskhou (commented):
Looks good to me! Two small  suggestions, and just waiting on the manual evicence when you post it. If we could run a startup, make sure the startup sequence works gerat and post in the PR what the expect telem would look like for reference i think this is about ready to merge

hrfarmer:
## On startup with `startup.seq` uploaded
https://github.com/user-attachments/assets/2ed69c1e-673d-4f4e-b175-60a0e07949a1

## Once `startup.seq` enables the burnwire
Screenshot 2026-07-27 at 13 25 55

## After reboot post-burnwire trigger
Screenshot 2026-07-27 at 13 26 47

## File size after `startup.seq` finishes and disables recording
image

PROVESFlightControllerReference/Components/TlmArchive/TlmArchive.cpp
```diff
@@ -0,0 +1,158 @@
...
+        }
+        {
+            Os::ScopeLock lock(this->m_queueMutex);
+            this->m_fileSize = currentSize;
+            this->m_fileSizeInitialized = true;
+        }
+    }
+
+    if ((currentSize > MAX_FILE_SIZE)) {
+        {
```
ineskhou: Its a soft check so no biggie but a currentSize + requested_size > MAX_FILE_SIZE is technically more correct

PROVESFlightControllerReference/Components/TlmArchive/TlmArchive.cpp
```diff
@@ -0,0 +1,158 @@
...
+TlmArchive ::TlmArchive(const char* const compName) : TlmArchiveComponentBase(compName) {}
+
+TlmArchive ::~TlmArchive() {}
+
+void TlmArchive::comIn_handler(FwIndexType portNum, Fw::ComBuffer& data, U32 context) {
+    (void)portNum;
+    (void)context;
+
+    {
+        Os::ScopeLock lock(this->m_queueMutex);
```
ineskhou, line 95: I dont think you need a mutex here, you are just emmiting an event, not writing to anything

PROVESFlightControllerReference/Components/TlmArchive/TlmArchive.cpp
hrfarmer: yeah thats how it was, but then if that check is there the component never actually hits the file size limit and will just emit the less precise failure count error instead

PROVESFlightControllerReference/Components/TlmArchive/TlmArchive.cpp
hrfarmer: i guess technically the events don't need to be in the mutex scope, but the mutex is for the variable assignments below it

PROVESFlightControllerReference/Components/TlmArchive/TlmArchive.cpp
ineskhou: I guess you could do a latch for that, which could be cleaner. Since the size isnt hard requirement this is more of a suggestion

PROVESFlightControllerReference/Components/TlmArchive/TlmArchive.cpp
ineskhou: Thoughts on using atomics instead? I think less mutex scope is best, especially when its on a rate group and a port

ineskhou (approved):
Thanks for reposnding to my comments! LGTM

ineskhou (changes requested):
@hrfarmer , actually I have 2 questions specifically about downlinking the file. I am having a hard time getting the physical tlm data so I can read it. Its through UART but I think it might be related to my GDS configuration. Do you mind downlinking and posting what a file would look like? I'm also wondering why you picked .tlm file, when a .txt might be more accesible?

ineskhou:
Did Testing with starting without the startup sequence.bin

The sat ran and
Screenshot 2026-08-02 at 4 07 16 PM

timed out
Screenshot 2026-08-02 at 4 07 36 PM

ineskhou:
Using The script was able unpack telelmtry

[tlm.txt](https://github.com/user-attachments/files/30640291/tlm.txt)

here is the original
[hello_tlm.csv](https://github.com/user-attachments/files/30640294/hello_tlm.csv)

ineskhou (dismissed):
Changed Paramaters, watched the buffer fill up based on the changed paramater, LGTM! Don't forget to post your evidence and merge once you've fixed the CI machine

hrfarmer:
> Changed Paramaters, watched the buffer fill up based on the changed paramater, LGTM! Don't forget to post your evidence and merge once you've fixed the CI machine

i'm testing it out right now to be sure, but the ci failure actually could be because of the component. this starts doing file system stuff right as the ci script tries to format the file system, so it could be causing some race condition/conflict that's leading to broken behavior. would explain why it seemed to sometimes work and sometimes not

hrfarmer:
At this point after commenting out the run handler in https://github.com/Open-Source-Space-Foundation/proves-core-reference/pull/464/commits/21309ec412cdb759e633b933d29beb183edcd991 and having the radio tests immediately start working again, I'm pretty convinced the component's somehow having an adverse effect on comms, rate groups, or the filesystem thats leading to upstream breakage. Given an issue like this could be a potential game-ender for the satellites and there's only a couple of days until handoff, unless anyone feel strongly otherwise (or it turns out to somehow be a simple fix) I don't feel confident merging for this release ☹️

cc @ineskhou

PR #439, opened 2026-07-11
feat: UHF radio port to Semtech USP — v5e board, UspRadio integration, carried patches (central tracking PR)

Mikefly123:
## AWAITING HUMAN REVIEW
This PR was generated by AI and is awaiting human cleanup before formal review and merge!

## Summary

Ports the UHF radio path from the Zephyr `drivers/lora` SX127x stack to Semtech's USP (LoRa Basics Modem RAL/RAC) for the SX1262 on PROVES FCB v5e. **This is the central PR for the USP port** — companion PRs:

- Open-Source-Space-Foundation/fprime-zephyr#14 — the `Zephyr::UspRadio` component (merge first)
- Open-Source-Space-Foundation/ground-radio-controller#25 — GRC-side port (merge after; re-point its submodule to fprime-zephyr main)

### What's here

- **FCB v5e board** (`boards/bronco_space/proves_flight_control_board_v5e/`): RP2350A + SX1262 (E22-400M30S: DIO3 TCXO 1.8 V / 10 ms, LDO reg-mode, external RF-switch GPIOs), DTS + defconfig + Kconfig.
- **DT-selected radio topology**: `RadioTopology_{Lora,Usp}.fppi` variants — v5/v5c/v5d keep the legacy LoRa component untouched; v5e builds UspRadio. Command surface preserved (TRANSMIT, CONTINUOUS_WAVE) + new SET_TX_PROFILE / SET_RX_PROFILE.
- **West manifest + carried patches** (`patches/`, applied via `make usp-patches` / `make usp-core-patches` / `make submodules`):
  - 0001 SX126x external RF-switch GPIO support; 0002/0003 Zephyr 4.3 build fixes
  - 0004/0005/0007 CDC-ACM robustness (incl. bounded poll_out backpressure — console-storm livelock fix)
  - 0006 SX126x wakeup busy-race settle delay
  - **0008 RAC api-mutex** (bare-metal no-op `protect_api_call` → k_mutex): torn radio-planner task slots
  - **0009 radio-planner failsafe UNLOCK exemption**: deterministic panic at unlock of any lock held >128 s
  - fprime: ComAggregator bounded-timeout (issue #432) + sched-tick-drop for rate-group-fed Svc components
- **ADRs 0001/0002** (USP-for-SX126x decision, link-profile table), `CONTEXT.md` flight-radio glossary, FatalHandler direct `sys_reboot`, FlashWorker guard, host UT wiring.

[...]

PR #421, opened 2026-07-10
Upgrade Zephyr to 4.4.1

nateinaction:
## Description

Upgrades the Zephyr RTOS from v4.3.0 to **v4.4.1**, building on the work in #365.

### Changes
- `west.yml`: bump `zephyr` to `v4.4.1` and align all module revisions (cmsis_6, hal_rpi_pico, hal_st, picolibc, mbedtls) to the v4.4.1 manifest.
- Replace `tinycrypt` (removed upstream) with `tf-psa-crypto` and drop it from the import allowlist — reflects the mbedtls crypto split in 4.4.
- Bump `mcuboot` to v2.4.0.
- Advance the `lib/zephyr-workspace/zephyr` submodule pointer to v4.4.1.
- `Main.cpp`: `FIXED_PARTITION_ID` → `PARTITION_ID` (renamed API).
- v5 devicetree: remove `compatible = "ti,tca9548a-channel"` and `label` from the tca954x mux channels — the 4.4 child-binding no longer declares them.
- v5 pinctrl: drop `ADC_CH0_P26` (GP26 is the tca9548a reset GPIO).

### Testing
- [x] `make build` — passes; produces `bootable.uf2` / `bootable.signed.hex` (Zephyr 4.4.1, board v5e/rp2350a/m33, SDK 1.0.1).
- [x] `make test-unit` — 4/4 pass.
- [x] `make fmt` (lint) — all pre-commit hooks pass.
- [x] UART integration tests on real v5e hardware — 26 passed / 6 skipped. Remaining failures are all unrelated to the upgrade: hardware not connected on the bench (burnwire, face sensors, power monitor, watchdog reset, radio RF path) or a pre-existing test-tooling issue (rtc `pytest.approx` datetime tolerance under newer pytest).

PR #466, opened 2026-07-22
ci: restore build-job reset step (dropped by #421), sparing the west workspace clones

Mikefly123:
## Summary

- **Restores the latent-state protection from #441**, which was accidentally removed from `main` by the Zephyr 4.4.1 upgrade (#421, commit 606eb0a) — since July 19 the self-hosted build job has had no reset step at all, re-exposing the patched-submodule state leakage #441 fixed.
- **Spares the west-managed Zephyr workspace from the top-level `git clean -dfX`** (`-e lib/zephyr-workspace`). The gitignored west module clones (zephyr, cmsis, hal_rpi_pico, mbedtls, …) are never tracked content; wiping them forces a full re-fetch on every build.
- **Resets each west module's working tree in place instead** (`git reset --hard && git clean -fdx` per clone). Carried patches — `patches/*.patch` today, plus the usp_zephyr/usp/zephyr patch sets landing in #439 — are applied into these trees, so they must start pristine every run or `west update` revision bumps land on dirty trees and the patch targets fail with 'Cannot apply'. Resetting the working tree gives the pristine guarantee while keeping the git object stores, so `west update` stays an incremental fetch (~86 s floor) instead of a re-clone (~4–5 min), even across `west.yml` revision bumps like 4.3 → 4.4.1.

Observed 'Setup Zephyr' timings on deathstar: 86–89 s steady-state incremental, 239–300 s when the runner crosses `west.yml` changes (4.4.1 upgrade, #439's new USP modules). This change keeps the job at the incremental floor while restoring run-to-run hygiene.

## Test plan

- [ ] CI build job green on this PR (first run may pay a one-time fetch)
- [ ] Re-run the build job to confirm 'Setup Zephyr' stays at the ~86 s incremental floor
- [ ] Verify `git status` inside `lib/fprime` and the west modules is clean at the start of the run (patch state from the previous run is reverted)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -45,6 +45,26 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
+      - name: Reset repository and submodules to original state
+        # Restores the latent-state protection from #441 (accidentally dropped
+        # by the Zephyr 4.4.1 upgrade in #421). Unlike the original step, the
+        # west-managed Zephyr workspace is excluded from the top-level clean and
+        # instead has each module's working tree reset in place: carried
+        # patches (patches/*.patch) are applied into these trees, so they must
+        # start pristine every run, but deleting the clones would force a full
+        # re-fetch of Zephyr and all modules on every build.
+        run: |
+          git submodule update --init --recursive
+          git clean -dfX -e lib/zephyr-workspace
+          git reset --hard
+          git submodule foreach --recursive git clean -dfX
+          git submodule foreach --recursive git reset --hard
+          find lib/zephyr-workspace -mindepth 1 -maxdepth 4 -name .git 2>/dev/null | while read -r g; do
+            d=$(dirname "$g")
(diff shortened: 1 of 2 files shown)


hrfarmer (dismissed):
lgtm (also added a note in `AGENTS.md` for it to not add the giant comment blocks they seem to love adding to the ci file)

hrfarmer (approved):

nateinaction (changes requested):
The submodule reset wasn't accidentally removed. Instead it was moved into the `make submodule` command. It's OK if we want to do this in CI but to prevent errors in other parts of the build we need to perform these actions before every call to `make submodules` hence why I moved it.

I think some of these actions are more than needed. The change to `make submodule` covers only the breaks we've seen in tests where the submodules have been modified and not fully reset in the next test run. I don't think we've seen breaks in tests related to all the additional calls in this step.

How do we think this will improve build time?

Mikefly123:
Closing this as potentially unneeded.

Issue #220, opened 2025-12-18
[BUG] Secret compiled into UF2 is extractable


nateinaction:
### Short summary

We publish UF2s as part of our builds, UF2s can be decompiled and authenticate secret can be extracted

### Steps to reproduce

n/a

### Expected behavior

Stop compiling the secret into the UF2 and instead configure the secret via param sent to satellite with initial configuration.

### Actual behavior

n/a

### Target board

RP2350 (Pico 2)

Mikefly123:
This might be fixed when we move to the Gen 3 Semtech LoRa IP and use an alternative method of doing message authentication.

We could also fix it by putting it in the parameter database.

We'll need to discuss how to operational solve this some more!

nateinaction:
```sh
Usage:
./extract_key_from_uf2.py bootable.uf2                # scan for candidates
./extract_key_from_uf2.py bootable.uf2 abc123   # check a specific key
```

```python
#!/usr/bin/env python3
"""Reassemble a UF2 image's data payload and search it for AUTH_DEFAULT_KEY.

UF2 splits firmware into 512-byte blocks (32 bytes of header/footer + up to
476 bytes of payload each), so a 32-character ASCII key can straddle a block
boundary and be invisible to plain `strings`. This script strips the framing
and searches the reassembled payload instead.
"""

import argparse
import re
import struct
import sys
from pathlib import Path

UF2_MAGIC_START0 = 0x0A324655
UF2_MAGIC_START1 = 0x9E5D5157
UF2_MAGIC_END = 0x0AB16F30

HEX32_NULL_TERM = re.compile(rb"[0-9a-fA-F]{32}\x00")


def reassemble_payload(uf2_bytes: bytes) -> bytearray:
    if len(uf2_bytes) % 512 != 0:
        print(
            f"Warning: input size is not a multiple of 512 bytes; "
            "it may not be a valid UF2 file.",
            file=sys.stderr,
        )

    payload = bytearray()
    for offset in range(0, len(uf2_bytes) - 511, 512):
        block = uf2_bytes[offset : offset + 512]
        magic0, magic1, _flags, _addr, payload_size, _block_no, _num_blocks, _family_id = (
            struct.unpack("<8I", block[:32])
        )
        (magic_end,) = struct.unpack("<I", block[508:512])
        if magic0 != UF2_MAGIC_START0 or magic1 != UF2_MAGIC_START1 or magic_end != UF2_MAGIC_END:
[...]

PR #472, opened 2026-07-23
Move HMAC auth key off firmware image onto internal-flash key store

nateinaction:
In this PR the command-authentication key moves off the firmware image and onto a dedicated littlefs partition on internal flash, mounted at `/keys`. The key was previously compiled into the binary as `AUTH_DEFAULT_KEY`, which issue #220 showed is trivially extractable from a shipped image, so baking it in defeated the point of authentication. The anti-replay sequence-number file moves to the same partition, since it previously lived on the unreliable SD-card FAT filesystem.

A board now boots keyless and is provisioned afterward. `TcSecurityDeframer` gains three commands: `PROVISION_KEY` for bootstrap (bypass-allowlisted so it works with no key on board, but only honored while the store is empty), and `ADD_KEY`/`REMOVE_KEY` for rotation (both require an authenticated frame). Up to 2 keys can be active at once so a rotation can add the new key before removing the old one. SPI validation now checks the active key store instead of a hard-coded SPI 0. The store is shared across all deframer instances (UART/LoRa/Sband): on an unknown SPI a deframer reloads the store from disk once before rejecting, so a rotation issued over one link is picked up by the others.

The ground-side framing plugin no longer reads a compiled-in default key; it takes `--authentication-key` or the `PROVES_AUTH_KEY` env var. CI provisions the board after flashing via a new `provision_key` integration test before running the authenticated integration suites.

Along the way, `Validator.cpp` had started depending on the FPP-generated `AuthKeyStore` type (which pulls in `Fw::Serializable`), breaking the project's pure-C++ contract for the `test-unit` gtest suite. It's decoupled via a small plain `ActiveSpiSlots` type in `Types.hpp`, with `TcSecurityDeframer::activeSpiSlots()` doing the conversion. New unit tests cover `parseHexKey`, `importHmacKeyBytes`/`destroyHmacKey`, and SPI/sequence-number validation against the new active-key-set shape.

## Test plan
- [x] `make test-unit` — all 8 suites pass, including expanded Authenticator/Validator coverage
- [x] `make generate build` — clean full build with no `AuthDefaultKey.h`, FLASH 69.82%, RAM 62.32%
- [x] `make check-console-disabled` — OK
- [x] CI green on this PR


PROVESFlightControllerReference/Components/ProvesRouter/Bypasser.cpp
```diff
@@ -44,6 +44,9 @@ static constexpr uint32_t kBypassOpCodes[] = {
     0x2100B000,  //!< ComCcsdsUart.tcSecurityDeframer.GET_SEQ_NUM
     0x2200B000,  //!< ComCcsdsLora.tcSecurityDeframer.GET_SEQ_NUM
     0x2300B000,  //!< ComCcsdsSband.tcSecurityDeframer.GET_SEQ_NUM
+    0x2100B002,  //!< ComCcsdsUart.tcSecurityDeframer.PROVISION_KEY
+    0x2200B002,  //!< ComCcsdsLora.tcSecurityDeframer.PROVISION_KEY
+    0x2300B002,  //!< ComCcsdsSband.tcSecurityDeframer.PROVISION_KEY
```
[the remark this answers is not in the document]
nateinaction: Agreed, there has to be a better way to handle these values. Make an issue for this.

README.md
```diff
@@ -152,7 +152,15 @@ MCUBoot only boots images that are **signed with the same key** the bootloader i
...
+You also want to make sure the authentication key the gds runs with is the same as the authentication key provisioned on the board. The board's key lives in its on-flash key store (never in the image); ground reads its key from the `--authentication-key` CLI arg or the `PROVES_AUTH_KEY` env var. Make sure these match the key you provisioned with `PROVISION_KEY`/`ADD_KEY`.
+
+The board holds up to two active keys so a rotation never leaves you locked out. Ground uses exactly one key at a time (whichever `--authentication-key`/`PROVES_AUTH_KEY` it was started with), so rotate in this order:
+
+1. Keep running GDS with the **old** key and send `ADD_KEY(new_spi, new_key)` — the command itself has to authenticate under the old key.
+2. Restart GDS with the **new** key (and `--spi new_spi`), and confirm commands are accepted.
+3. Only then send `REMOVE_KEY(old_spi)`, authenticated under the new key.
+
+Doing step 3 before step 2 works too, but leaves nothing to fall back on if the new key turns out to be wrong. `REMOVE_KEY` refuses to remove the last remaining key.
```
ineskhou, line 171: Would be helpful if the readme was more clear about how to initally provision a key. Maybe its somewhere that I havent read yet, but ideally you could read the readme from start to finish as a begining person and be able to run authenticated

Makefile
ineskhou: Would be great to include  --spi here so we can use make gds to run other keys

boards/bronco_space/proves_flight_control_board_v5/proves_flight_control_board_v5.dtsi
```diff
@@ -23,6 +23,17 @@
...
+                        compatible = "zephyr,fstab,littlefs";
+                        mount-point = "/keys";
+                        partition = <&keystore_partition>;
+                        automount;
+                        read-size = <16>;
+                        prog-size = <16>;
+                        cache-size = <64>;
+                        lookahead-size = <32>;
+                        block-cycles = <512>;
+                };
```
ineskhou, line 36: Michael got some help for the review of risks https://claude.ai/code/session_01VRCjXtQCFyoBvekU4zcCQX 

it pointed out that  if Zephyr fails to mount the littlefs volume at /keys, it mounts the empty one. I think this is the desired behvaour, a empty filesytem allows for recovery unlike a corrupt one. However it would allow someone to provision a first key which could be seen.

We have observed FAT filesystems failing in the sd card, but this is a littlefs. We could move the sequence number off here and back to the regular filesusyem to hammer the filesystem better. Could also think of a better wat to fall bacm on a failed filesysten

I'm highlighting this one bc of the suspected filesystem issues we've had on the last 3 sats (unconfirmed)

PROVESFlightControllerReference/Components/TcSecurityDeframer/TcSecurityDeframer.hpp
```diff
@@ -97,6 +118,43 @@ class TcSecurityDeframer final : public TcSecurityDeframerComponentBase {
...
+    //! Finds the PSA key id for the given SPI among currently-imported keys.
+    //! Must be called with the key store lock held.
+    bool findKeyIdForSpi(uint32_t spi, uint32_t& keyId) const;
+
+    //! Returns true if any valid slot already holds the given SPI.
+    //! Must be called with the key store lock held.
+    bool hasSpi(uint16_t spi) const;
+
+    //! Returns the number of valid slots in m_keyStore. Must be called with the key store lock held.
+    U8 activeKeyCount() const;
```
ineskhou, line 152: Could be good to get this, as well as what spis exist, in case we forgot what slot we put keys in? not super important tho

docs-site/components/TcSecurityDeframer.md
```diff
@@ -9,10 +9,26 @@ The component does not enforce policy. Frames that fail verification are still f
...
+
+`REMOVE_KEY` zeroizes the revoked slot's key bytes before persisting, so a removed key is not recoverable from the on-flash record; the slot is restored if the durable write fails.
```
ineskhou, line 32: Thank you for writing this all out!! Really cool design + made me think a lot about risk miti=gation

Framing/src/authenticate_plugin.py
```diff
@@ -19,54 +19,67 @@
...
+
+
+def get_auth_key_from_env() -> str:
+    """
+    Read the authentication key from the PROVES_AUTH_KEY environment variable.
+
+    The key is never compiled into the flight image (see issue #220), so ground
+    tooling must be told the key out-of-band: via --authentication-key or this
+    environment variable.
+
```
ineskhou, line 67: this would be good 2 hav in readme

ineskhou:
Confirmed

- ✅ This branch did not move MCUBoot’s image slots. Compared to main:

Testing
-over the air update
- works with one key provisioned
Screenshot 2026-08-02 at 11 51 58 PM
- works with two keys provisioned 

- refuses with wrong key 
Screenshot 2026-08-02 at 11 53 51 PM

-Watchdog / power-cut mid-command storm


Note:

- Lora and UART sequence numbers are diff, rmember to write down correct sequence number at checkouts
- FsFormat then check ActiveKeyCount (diff systems should befine_

PROVESFlightControllerReference/Components/TcSecurityDeframer/TcSecurityDeframer.hpp
nateinaction: I agree, it would be good to have an SPI list where we get fingerprints back so we can ID which keys are onboard.

boards/bronco_space/proves_flight_control_board_v5/proves_flight_control_board_v5.dtsi
nateinaction: Hmmm based on my reading of this comment it sounds like we might want to move the seq num back only to experiment trying to understand FAT failure modes? This doesn't sound like a request for this PR where we want to introduce a reliable component.

Callout for mounting risk sounds OK to me. Good callout.

Let me know if I misread this.

README.md
nateinaction: Added a walkthrough of the first-time PROVISION_KEY bootstrap in a04d469cf36a6a06c0fa96f29e0c6d5c7c231c9f.

Makefile
nateinaction: Added SPI=<n> support to make gds in 8d2a5aed.

Framing/src/authenticate_plugin.py
nateinaction: Folded this into the new README walkthrough in a04d469cf36a6a06c0fa96f29e0c6d5c7c231c9f.

PROVESFlightControllerReference/Components/TcSecurityDeframer/TcSecurityDeframer.hpp
nateinaction: Filed https://github.com/Open-Source-Space-Foundation/proves-core-reference/issues/488 to track this — a bit more involved than fits in this PR (new command + fingerprint, not just exposing existing state).

PR #489, opened 2026-08-03
DRAFT: MosaicManager integration test run (TXST MOSAIC payload)

nateinaction:
## Summary

Adds `MosaicManager`, a passive F Prime component that records data from the MOSAIC gamma ray payload to the filesystem under `/mosaic/` for later downlink.

This draft exists to **run the integration tests against the MOSAIC hardware on the OSSF integration runner**. The component itself is destined for the Texas State University fork, which is flying the MOSAIC payload.

## What it does

MOSAIC streams ASCII CSV lines of the form `ADC=<raw>,MV=<millivolts>\n` at 9600 baud, one sample every 100 ms, over the second peripheral UART (`uart1`). `MosaicManager` only listens — it never sends commands to the payload, so the UART driver connects straight to the manager with no `PayloadCom` in between.

Parsed samples are appended as fixed-size 8-byte binary records (`U32` seconds time tag, `U16` ADC, `U16` millivolts) to files named `/mosaic/gamma_<seq>.dat`. A file closes when it fills (100 samples), when a partially filled file goes stale (60 s, checked on the 1 Hz rate group), on `FLUSH`, or on `STOP_RECORDING`.

Commands: `START_RECORDING`, `STOP_RECORDING`, `FLUSH`. Telemetry: `Recording`, `SamplesRecorded`, `FilesWritten`, `ParseErrors`, `LatestAdc`, `LatestMillivolts`, packetized as `Mosaic` (packet id 23, group 4).

## Notes on the approach

The first cut of this used F Prime data products. The `DpCatalog` did not fit in memory on the rp2350, so samples are written directly through `Os::File` instead — hence the branch name.

`ensureFileOpen()` probes forward with `Os::FileSystem::getPathType()` for the first unused file index before opening. The sequence counter lives in RAM and resets to 0 on reboot, and the Zephyr `Os::File` delegate truncates on `OPEN_CREATE` regardless of `NO_OVERWRITE` (its `overwrite` handling is unimplemented). Without the probe, a reboot would silently wipe a file from a previous boot that has not been downlinked yet.

## Supporting changes

- `prj.conf`: `CONFIG_FS_FATFS_REENTRANT` (plus `LFN_MODE_STACK`, which it depends on) so concurrent FatFs access is safe, and `CONFIG_ZVFS_OPEN_ADD_SIZE_POSIX=16` — the POSIX fd/DIR pool defaults to 3, sized for stdio, which is not enough with `MosaicManager`, `FileManager`, and `FileDownlink` all holding handles.
- `PlatformCfg.fpp`: `FW_DIRECTORY_HANDLE_MAX_SIZE` 16 → 272, since `ZephyrDirectory` stores a 256-byte path for `rewind()` and the topology now creates `/mosaic` at startup.
- `TlmPacketizerCfg.hpp`: `MAX_PACKETIZER_CHANNELS` 202 → 220 to cover the new channels.

## Integration tests

`PROVESFlightControllerReference/test/int/mosaic_manager_test.py`, marked `uart_only`:

1. `START_RECORDING` / `STOP_RECORDING` round-trip, asserted through both the event and the `Recording` channel.
2. Samples stream in from the attached payload and `SamplesRecorded` advances, with `LatestAdc` / `LatestMillivolts` confirming the ASCII protocol was actually decoded rather than bytes merely counted.
3. `FLUSH` closes the file, and `FileHandling.fileManager.FileSize` reports exactly `records * 8` bytes for the path named in `SampleFileClosed` — end-to-end proof the samples reached the filesystem and are available for downlink.

## Not for merge

Draft, and intentionally so — this is not a contribution to upstream. It runs here because this is where the MOSAIC hardware is attached to the runner.

---

## Hardware test results

[integration-uart, run 30827650821](https://github.com/Open-Source-Space-Foundation/proves-core-reference/actions/runs/30827650821/job/91734527604) — `mosaic_manager_test.py ...` — all three MosaicManager tests pass against the MOSAIC payload attached to the runner. `integration-radio`, `build`, `unit-test`, `lint` and `yamcs-build` are all green.

The one red test in that run is `mode_manager_test.py::test_safe_09_command_loss_triggers_safe_mode_and_reboot`, which is unrelated to this branch and flaky on the runner — across the runs on this branch it failed, passed, passed, then failed again.

## Bugs this branch had to fix to get there

Recorded here because most of them are not MosaicManager-specific and are worth knowing about:

1. **The deployment was one command away from not booting.** `CMD_DISPATCHER_DISPATCH_TABLE_SIZE` was 350 and the deployment had 348 commands. Adding three took it to 351; the table filled during topology setup and the next registration tripped `FW_ASSERT` in `CommandDispatcherImpl::compCmdReg_handler`, so the board FATALed at boot and never answered a command. This is not a build error — any new component with a few commands would have hit it. Raised to 400.
2. **`CONFIG_FS_FATFS_REENTRANT` hard-faults the board.** It requires `!FS_FATFS_LFN_MODE_BSS`, so it had been paired with `LFN_MODE_STACK`. With `EXFAT` enabled the LFN working buffer is `(FS_FATFS_MAX_LFN + 1) * 2 + 608` bytes — about 1.1 KB at the default `MAX_LFN` of 255 — and STACK mode puts that on the stack of *every* thread that touches the filesystem, overflowing the rate group stacks. Reverted to main's non-reentrant configuration. Making the deployment's FatFs access genuinely thread-safe via `LFN_MODE_HEAP` is worth doing but is a global change that belongs in its own PR.
3. **Cross-thread races in the component.** `dataIn` runs on the 10 Hz rate group thread, `run` on the 1 Hz thread, and the commands on the command dispatcher thread; all three mutate the open file handle and its counters. As `sync` ports on a passive component they had no serialization at all. Now `guarded`.
4. `ReferenceDeploymentPackets.fppi` is evaluated inside the `ReferenceDeployment` module, so fully-qualified channel names are rejected by `fpp-to-dict`. Also dropped a `CdhCore.tlmSend.SendLevel` entry referencing a channel that is not defined anywhere in the tree.
5. Unresolved merge conflict markers in `TlmPacketizerCfg.hpp` from the merge from main.

## Notes for the OSSF maintainers

Two things this branch ran into that are not mine to change, but that you may want to:

- **`recover_from_safe_mode` only runs on the radio pass.** `mode_manager_test.py::test_safe_09` deliberately drives the FSW into SAFE_MODE and reboots. When it fails, its own teardown does not finish recovering, and on the UART pass conftest's fixture returns early — so the board stays in SAFE_MODE with the payload load switch held off, and every test in the next file fails for reasons unrelated to itself. `mosaic_manager_test.py` sorts immediately after `mode_manager_test.py` and was the first casualty. I worked around it locally by checking the mode in my fixture, but extending `recover_from_safe_mode` to cover UART would fix it for everyone.
- **Telemetry assertions need timeouts over 30 s.** `telemetryDelay` (`Utilities.RateDelay` on the 1 Hz rate group) gates `CdhCore.tlmSend.Run`, so no packet downlinks more often than every 30 s regardless of the packetizer's `ON_CHANGE_MIN`/`min=0`/`max=0` group config. Since `proves_send_and_assert_command` clears the histories before sending, any post-command telemetry assertion waits a full downlink period. No other integration test asserts on telemetry — they all use events, which are not gated — so nothing had hit this before. Might be worth a documented constant in `common.py`.


naa1992m-maker:
Variables

ALPACA_API_KEY=...
ALPACA_SECRET_KEY=...
ALPACA_PAPER=true

Issue #429, opened 2026-07-10
[FEATURE] OTA Improvements and robust Testing


ineskhou:
### Short summary

Make over-the-air firmware updates faster and more reliable. We cannot do true binary patching with MCUBoot, but we can speed up updates by keeping a flight image on the filesystem and improving the update workflow and test coverage.

### Motivation and user story

OTA updates take a long time over a short pass. 

### Proposed implementation

- Improve operator experience
- look into making the process faster approching a patch compoent

### Priority

Medium

yudataguy:
## OTA Improvements — design proposal

Analysis of the current OTA path with measurements taken from real CI build artifacts, five defects found in the existing implementation, and a phased plan. Open questions from the first review round are resolved in section 5.

# OTA Improvements — Design Proposal (#429)

Measured analysis of the current OTA path, five defects found in it, and a phased plan.
All measurements below are from real CI build artifacts, not estimates.

---

## TL;DR

| | bytes to uplink | time @ 510 B/s | fits one pass? |
|---|---|---|---|
| **Today** — full signed image | 726,784 | **23.8 min** | no |
| gzip'd image | 378,362 | 12.4 min | no |
| **bsdiff+heatshrink delta** ← selected | ~97,000 | **3.2 min** | yes |
| bsdiff+LZMA delta | ~47,000 | 1.5 min | yes |

Delta patching is worth **7–15×** on uplink time and is the difference between "cannot
be done in a pass" and "comfortably done in a pass". Naive block-level diffing is *not* —
it is measured below at 0% savings.

Separately, the current `FlashWorker` **reports failed updates as successful**
(`FlashWorker.cpp:64-69`). That is a correctness bug that should be fixed independently of
any of the performance work.

---

## 1. How OTA works today

Two pieces:

- `lib/fprime-extras/FprimeExtras/Update/` — generic `Updater` component and the
  `UpdateWorker` port interface (`Ports/Ports.fpp:47`). **Library code, out of scope per
  the decision to keep this work PROVES-side.**
- `PROVESFlightControllerReference/Components/FlashWorker/` — the PROVES worker, bound in
  place of the library's `StubWorker` by the config override at
  `project/config/UpdateConfig.fpp:29`.

Operator sequence (from `test/manual_stress_test.py:230`):

```
uplink /update/zephyr.signed.bin      (FileUplink, ~24 min)
  → PREPARE_UPDATE                    (erase 1 MB slot)
[...]

PR #503, opened 2026-08-05
feat(ota): fix update failure reporting, add progress telemetry and delta patching

yudataguy:
Implements [#429](https://github.com/Open-Source-Space-Foundation/proves-core-reference/issues/429). Design proposal and measurements were posted [on the issue](https://github.com/Open-Source-Space-Foundation/proves-core-reference/issues/429#issuecomment-5186975856) and agreed before this was written.

## The problem, measured

A signed image is **726,784 bytes**. The ground station paces file uplink at 204-byte chunks with a 0.4 s cooldown (`fprime-gds.yml`), about **510 B/s**, so a whole image needs **~24 minutes** of contact. A pass is single-digit minutes, and F Prime's file uplink has no cross-pass resume, so a dropped pass discards everything sent.

## Correctness (the part worth reviewing closely)

`FlashWorker::writeImage` only ever assigned its return status in the CRC-mismatch branch. A file read error or flash write error logged a warning and returned `OP_OK`, so `Updater` emitted **`UpdateSucceeded` for an image that was never fully written** — and an operator following the documented procedure would then set next-boot and reboot into a truncated image.

The write path is restructured so every exit returns a classified `WriteOutcome`; there is no longer a mutable status a failure path can forget to set. Also fixed: the CRC-mismatch path fell through into the write loop instead of returning, and *any* failure forced another full 1 MB erase even when flash was never touched.

Failures that never reached flash now leave the slot usable, so a mistyped filename costs a retry instead of an erase. Failures that did reach flash still require another prepare.

## Operator visibility

`UpdateStage`, `BytesWritten`, `ImageTotalBytes`, `LastUpdateStatus`, packetized as `SoftwareUpdate` (23, group 5). Channels rather than only events, so an operator returning a pass later can see where an update stands without replaying event history. `CHUNK_DELAY_US` and `PROGRESS_STEP_PERCENT` are parameters so write pacing can be tuned against hardware rather than rebuilt.

## Fitting an update into a pass

**`ASSEMBLE_IMAGE`** joins numbered segments (`<prefix>.000`, ...) uplinked across several passes and verifies the result, bounding the loss from a dropped pass to one segment.

**`APPLY_PATCH`** reconstructs an image from a bsdiff-style delta against the running image, read straight out of `slot0` — the RP2350 executes XIP from memory-mapped QSPI, so no filesystem copy is needed (this is simpler than the issue's "keep an image on the filesystem" suggestion, and removes the stale-copy failure mode).

Measured against real consecutive CI builds:

| reference age | delta | uplink |
|---|---|---|
| 2 days | ~47 KB (LZMA) | 1.5 min |
| 12 days | ~66 KB (bzip2) | 2.1 min |
| whole image | 726,784 B | 23.8 min |

Naive block diffing was measured and **rejected**: a code-size change shifts every later address, so ~99% of 512-byte blocks differ between builds days apart — a block-index scheme would uplink *more* than the image.

## Safety of the patch path

The container records the **size and CRC32 of the reference it was built from**, and the apply refuses to run against anything else. Patching the wrong reference yields a plausible but corrupt image that would then be flashed and booted. Control records are range-checked against both images, so a malformed patch arriving over the radio cannot read or write out of bounds.

## One decision left to the team

**Compression is deliberately not implemented on the flight side.** The patch streams are ~84% zero bytes and compress from 727 KB to ~60 KB (DEFLATE), ~47 KB (LZMA), or ~97 KB (heatshrink) — but this Zephyr workspace ships no decompressor, and adding one to flight software is a dependency decision I did not want to make unilaterally. I also measured a naive zero-RLE (no new dependency): only 241 KB / 7.9 min, because the zeros come in short scattered runs. A real LZ-class coder is required.

The container carries a codec field so a compressed format drops in without changing the applier's structure, and `make-patch.py` refuses to emit an uncompressed patch without `--allow-uncompressed`, since it would be the size of the image. **Until a codec is chosen, `APPLY_PATCH` is functional but not yet operationally useful** — `ASSEMBLE_IMAGE` is the shipping answer to the pass-boundary problem.

## Also

Partition IDs come from the device tree via `FIXED_PARTITION_ID` instead of a hand-counted literal that depended on DTS declaration order. Partition labels corrected — `slot1` was labelled `"golden"` while being the slot OTA overwrites.

## Verification

- `make fmt` — clean
- `make test-unit` — 11/11 pass, three new host suites (`UpdateSequencer`, `SegmentPlan`, `PatchApplier`) covering the defects above plus malformed and hostile patches
- Firmware builds; `zephyr.uf2` produced
- **End to end**: `make-patch.py` plus the flight-side applier reconstructed a real 726,784-byte image from a real 2-day-older reference, **byte-identical**
- `test/int/ota_test.py` added for the command surface. It deliberately never sets next-boot, so a run cannot leave a board staged to boot an unintended image. **Not yet run against hardware.**

Not hardware-tested: the full prepare → write → TEST boot → confirm cycle, and `APPLY_PATCH` on a real board.

Note for reviewers: `lib/fprime-extras` in my working tree was ahead of the recorded pointer and declares `Svc.ComRetry`, which collides with `lib/fprime`'s. I pinned it to the recorded commit to build, then restored it; this branch does not touch the submodule.

https://claude.ai/code/session_011kehXMi9zAAoPJ3PgfRd81

### DIFF
--- a/PROVESFlightControllerReference/Components/FlashWorker/CMakeLists.txt
+++ b/PROVESFlightControllerReference/Components/FlashWorker/CMakeLists.txt
@@ -20,6 +20,9 @@ register_fprime_library(
         "${CMAKE_CURRENT_LIST_DIR}/FlashWorker.fpp"
     SOURCES
         "${CMAKE_CURRENT_LIST_DIR}/FlashWorker.cpp"
+        "${CMAKE_CURRENT_LIST_DIR}/UpdateSequencer.cpp"
+        "${CMAKE_CURRENT_LIST_DIR}/SegmentPlan.cpp"
+        "${CMAKE_CURRENT_LIST_DIR}/PatchApplier.cpp"
 #   DEPENDS
 #       MyPackage_MyOtherModule
 )
--- a/PROVESFlightControllerReference/Components/FlashWorker/CMakeLists.txt
+++ b/PROVESFlightControllerReference/Components/FlashWorker/CMakeLists.txt
@@ -23,6 +23,7 @@ register_fprime_library(
         "${CMAKE_CURRENT_LIST_DIR}/UpdateSequencer.cpp"
         "${CMAKE_CURRENT_LIST_DIR}/SegmentPlan.cpp"
         "${CMAKE_CURRENT_LIST_DIR}/PatchApplier.cpp"
+        "${CMAKE_CURRENT_LIST_DIR}/LzssDecoder.cpp"
 #   DEPENDS
 #       MyPackage_MyOtherModule
 )
--- a/PROVESFlightControllerReference/Components/FlashWorker/FlashWorker.cpp
+++ b/PROVESFlightControllerReference/Components/FlashWorker/FlashWorker.cpp
@@ -8,6 +8,8 @@
 
 #include <Utils/Hash/libcrc/lib_crc.h>  // same CRC primitive Os::File::calculateCrc uses
 
+#include <limits>
+
 #include "Os/File.hpp"
 #include "Os/Task.hpp"
 #include <zephyr/dfu/flash_img.h>
@@ -75,7 +77,7 @@ static_assert(static_cast<U8>(UpdateSequencer::Stage::FAILED) == static_cast<U8>
 // ----------------------------------------------------------------------
 
 FlashWorker ::FlashWorker(const char* const compName)
-    : FlashWorkerComponentBase(compName), m_last_reported_percent(0) {}
+    : FlashWorkerComponentBase(compName), m_pending_confirm_seconds(0), m_last_reported_percent(0) {}
 
 FlashWorker ::~FlashWorker() {}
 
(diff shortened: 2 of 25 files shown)


yudataguy:
Both open decisions are now resolved and implemented in a29752b.

## 1. Compression: a self-contained LZ77 codec, not a vendored library

The choice was heatshrink vs. LZMA vs. DEFLATE, all of which meant adding a third-party decompressor to flight software. Measuring first changed the answer: a ~120-line LZ77 coder whose **exact decoder can be round-tripped against real patches in host tests** matches heatshrink's result without the dependency.

Measured on a real patch between consecutive CI builds:

| | size | uplink |
|---|---|---|
| raw patch | 728,388 B | 23.8 min |
| **LZ77, 4 KB window** | **101,208 B** | **3.3 min** |
| LZ77, 16 KB window | 95,496 B | 3.1 min |
| LZ77, 64 KB window | 93,814 B | 3.1 min |
| heatshrink (earlier measurement) | ~97,000 B | 3.2 min |
| naive zero-RLE | 240,921 B | 7.9 min |

**4 KB window.** A 16× larger window buys 0.2 minutes — the compressible content is bsdiff's difference stream, which is short zero runs repeating at close range, so distance is cheap and window size barely matters. Zero-RLE was measured and rejected: the zeros come in short scattered runs (~10 bytes), so a real LZ-class coder is required.

The three streams are compressed together and decoded once into a scratch file, so `PatchApplier` still operates on plain streams and its existing tests cover the apply unchanged.

**Verified end to end**: decoding a real 101,208-byte compressed patch and rebuilding a real 726,784-byte image from it, byte-identical to the target.

## 2. Auto-confirm: operator-armed, defaulting to off

To answer the question directly — **yes, the operator arms it from the ground, and it takes two commands**:

```
Update.worker.AUTO_CONFIRM_ENABLED_PRM_SET(TRUE)
Update.worker.AUTO_CONFIRM_ENABLED_PRM_SAVE
```

The `PRM_SAVE` is not optional. The confirm decision happens *after* the reboot into the test image, so an unsaved parameter would be lost exactly when it is needed. `FileHandling.prmDb` is the real `Svc.PrmDb` and its file lives on the LittleFS storage partition, which the slot swap does not touch, so an armed setting survives the update.

Confirmation is refused unless all three hold: it was armed from the ground, the running image is actually pending confirmation (`boot_is_img_confirmed()` is false), and it has run continuously for `AUTO_CONFIRM_DELAY_SECONDS` (default 1800). A configured delay of zero still requires one elapsed second, so an image that crashes on boot cannot confirm itself. `RunningImageConfirmed` and `PendingConfirmSeconds` are downlinked so the countdown is visible.

Default is `false` — confirmation stays operator-in-the-loop until deliberately handed over.

The check runs on the 1 Hz rate group. `ActiveRateGroupOutputPorts` is configured at 25 and member 19 was free, so no config change was needed.

## Verification

`make fmt` clean, `make test-unit` **12/12** (a new `LzssDecoder` suite covering overlapping matches, ring-buffer wrap, and malformed streams), firmware builds.

Still not hardware-tested: the full prepare → write → TEST boot → confirm cycle, `APPLY_PATCH` on a real board, and the auto-confirm path across an actual reboot.

PR #513, opened 2026-08-30
Move TC uplink security onto upstream F Prime SDLS interfaces

nateinaction:
## Summary
- Replace `Components::TcSecurityDeframer` (a project-local pass-through component written before F Prime had a security hook) with `Components::TcSecurityDecryptor`, a decryptor implementing `Svc.Ccsds.CcsdsSdlsDecrypt` behind the upstream `Svc::Ccsds::CcsdsSdlsDeframer` introduced in F Prime 4.3.0.
- The security association index is now supplied by the deframer instead of being parsed from the frame, so MAC verification moves to the PSA multipart API (`psa_mac_verify_setup`/`update`/`update`/`psa_mac_verify_finish`) to keep the SA index inside the authenticated span.
- Wire format, the `authenticated`-flag policy design, ground-side Framing plugin, and all command opcodes are unchanged. No `SdlsSaRouter` or `SdlsFileKeyManager` adopted — both rejections are documented in the component SDD, along with the seam #472 (key store) and #490 (key fingerprints) can land through.

## Test plan
- [x] `make test-unit` — all 8 host gtest suites pass (added coverage that the SA index participates in the MAC)
- [x] `make generate build` — clean full Zephyr flight build; `GET_SEQ_NUM` opcodes confirmed unchanged (`0x2100B000`/`0x2200B000`) and `CurrentSequenceNumber` channels present in the topology dictionary for UART/LoRa
- [x] `make check-console-disabled`
- [x] Bench integration on the flight control board: flashed, ran `sync_sequence_number` then the broader integration suite (11 passed, 0 failed); event log shows the anti-replay path organically rejecting a duplicate frame (`SequenceNumberInvalid`) without advancing the counter, no `AuthenticationFailed`/`ParsingFailed`/`DecryptionFailed` events, `commsBufferManager` high-water stable
- [x] LoRa-link round-trip and unauthenticated-bypass/truncated-frame scenarios — not exercised on this bench (no radio hardware attached, no raw-frame-injection tooling exists yet in the ground tooling; noted as a pre-existing follow-up gap)


PROVESFlightControllerReference/Components/TcSecurityDecryptor/TcSecurityDecryptor.cpp
```diff
@@ -22,56 +22,60 @@ namespace Components {
...
     const Ccsds355_0_B_2::TcTransferFrame::Parser::Result parseResult =
         Ccsds355_0_B_2::parse(data.getData(), data.getSize());
     if (parseResult.status != Ccsds355_0_B_2::TcTransferFrame::Parser::Status::Ok) {
         // The frame is too short to contain the security fields, so it cannot be stripped
-        // for downstream deframing. Return buffer ownership upstream and drop the frame.
+        // for downstream deframing. Report a decryption failure so the upstream deframer
+        // drops the frame and returns the buffer for deallocation.
         this->log_WARNING_HI_ParsingFailed(static_cast<PacketParserStatus::T>(parseResult.status));
-        this->dataReturnOut_out(0, data, contextOut);
+        this->decryptOut_out(0, Svc::Ccsds::SdlsStatus::DECRYPTION_FAILURE, data, contextOut);
```
nateinaction, line 55: **medium — unthrottled event amplification from the uplink**

Routing the parse-failure drop through `decryptOut_out(DECRYPTION_FAILURE, ...)` makes the upstream deframer emit `Svc.Ccsds.CcsdsSdlsDeframer.DecryptionFailed`, which is declared **with no throttle**, in addition to the local `ParsingFailed` (throttle 2).

Producing a droppable frame requires no secret — only a well-formed TC primary header and a correct FECF CRC16. So anyone transmitting on the LoRa/UART link can emit one unthrottled `WARNING_HI` event per frame indefinitely, crowding real events out of the ComQueue.

The interface offers a quieter drop: `bufferReturnOut` is documented as "Port for returning the incoming iv/data buffer for deallocation". Calling

```cpp
this->bufferReturnOut_out(0, data, contextOut);
```

reaches `CcsdsSdlsDeframer::bufferReturnIn_handler` -> `dataReturnOut_out` -> `TcDeframer`, returning the buffer with no upstream event.

PROVESFlightControllerReference/Components/TcSecurityDecryptor/TcSecurityDecryptor.fpp
```diff
@@ -8,18 +8,20 @@ module Components {
...
+    @ Decryptor behind the upstream Svc.Ccsds.CcsdsSdlsDeframer, implementing
+    @ Svc.Ccsds.CcsdsSdlsDecrypt. Implements the TC ProcessSecurity flow of CCSDS 355.0-B-2:
+    @ parse the Security Trailer, validate the security association index and anti-replay
+    @ sequence number, and verify the MAC (which covers the security association index,
+    @ supplied by the deframer, and the frame data). The verification result is recorded in
+    @ the frame context (authenticated flag); policy enforcement (reject or bypass) is owned
+    @ downstream by the router.
+    passive component TcSecurityDecryptor {
+
+        import Svc.Ccsds.CcsdsSdlsDecrypt
```
nateinaction, line 24: **low — recursive acquisition of the guarded-port mutex**

`import Svc.Ccsds.CcsdsSdlsDecrypt` makes both `decryptIn` and `decryptReturnIn` **guarded**; the generated `TcSecurityDecryptorComponentAc.cpp` (lines 1038 and 1066) confirms both call the same `this->lock()`.

The whole downstream chain is synchronous (`spacePacketDeframer` -> `ProvesRouter::dataIn_handler`, which calls `dataReturnOut_out` on the same stack), so `decryptReturnIn_handler` re-enters `m_guardedPortMutex` while `decryptIn_handler` still holds it. This only works because Zephyr's `k_mutex` is recursive — the removed component deliberately declared `sync input port dataReturnIn` to avoid exactly this.

On any POSIX host build, `Os::Posix::Mutex` uses `PTHREAD_MUTEX_ERRORCHECK` and would `FW_ASSERT`, which will block re-enabling the commented-out component UT in `CMakeLists.txt:52`. Worth an explicit comment at minimum, or overriding `lock()` / `unLock()`.

nateinaction (commented):
## Code review

Reviewed the rename and re-platforming onto `Svc.Ccsds.CcsdsSdlsDeframer` / `CcsdsSdlsDecrypt`. The substance is correct. Verified:

- **Wire format compatibility holds.** `authenticateFrame` MACs `saIndex(2, BE) || [SeqNum|Data]`, which matches `Framing/src/authenticate_plugin.py`'s `HMAC(SPI|SeqNum|Data)` exactly.
- **Test vectors are right.** Recomputed both new vectors in `test_TcSecurityDecryptor_Authenticator.cpp`.
- **Opcodes preserved.** `ReferenceDeploymentTopologyDictionary.json` shows `0x2100B000` / `0x2200B000` for `GET_SEQ_NUM`, keeping the `Bypasser.cpp` literals valid; `CurrentSequenceNumber` exists on both links.
- **Buffer ownership is sound** on both the SUCCESS and `DECRYPTION_FAILURE` paths. Dropping the un-slice in the return handler is safe: `BufferManager::bufferSendIn_handler` only asserts the pointer is inside the allocation and that the size did not grow.
- **Truncated MAC works in the multipart PSA path** — `psa_mac_finalize_alg_and_key_validation` only rejects `< 4` bytes, and the imported key's `PSA_KEY_USAGE_VERIFY_MESSAGE` is the flag `psa_mac_verify_setup` requires. No fallback needed.
- **No base-id collisions.** The new `sdlsDeframer` base id `+0x0C000` is clear in all three subtopologies; `logOut` / `logTextOut` / `timeCaller` are auto-connected, and the unconnected `errorNotify` is guarded upstream.

Three inline findings below, plus one that is not anchorable because the file is not in the diff:

### Stale debug configs — `.vscode/launch.json:38` (low)

Three debug configurations still point at `build-gtest/test_TcSecurityDeframer_{Authenticator,Parser,Validator}`. After this rename `make test-unit` produces `test_TcSecurityDecryptor_*`, so all three fail to launch. This file was missed by the rename fan-out.

PROVESFlightControllerReference/Components/TcSecurityDecryptor/Const.hpp
```diff
@@ -16,18 +16,16 @@ namespace Ccsds355_0_B_2 {
 //! Telecommand packet structure
 
 //! CCSDS 355.0-B-2 Section E2.2
+//! The Security Parameter Index is stripped by the upstream Svc.Ccsds.CcsdsSdlsDeframer before
+//! this component sees the frame; kSpiSize is kept only for the MAC-prefix fallback in Authenticator.
```
nateinaction, line 20: **low — comment refers to code that does not exist**

This says `kSpiSize` is "kept only for the MAC-prefix fallback in Authenticator", but no such fallback was implemented: `Authenticator.cpp:113` hardcodes `const uint8_t saIndexBytes[2]`. `kSpiSize` is now referenced nowhere, so the comment points a future reader at code that is not there.

